PR #4, opened 2023-08-04
Adding more detailed endpoints, a POJO model and response classes, gi…

Lukehub1:
…tignorefile


src/main/java/uk/gov/laa/pfla/auth/service/models/MappingTableModel.java
deepujames1234, line 1: Follow java naming convention, camelcase fields

PR #18, opened 2023-11-28
Lpf 202 dbc code to query static csv sql mapping table

Lukehub1:


src/main/java/uk/gov/laa/pfla/auth/service/dao/MappingTableDao.java
```diff
@@ -1,53 +1,116 @@
...
 import java.util.List;
+
 @Repository
-public class MappingTableDao {
+@Slf4j
+public class MappingTableDao  implements RowMapper<Object> {
 
     private final List<MappingTableModel> mappingTableObjectList = new ArrayList<>();
+
+//    @Autowired
```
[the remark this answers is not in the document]
Lukehub1: Yeh I could delete that now really... I went with a different syntax in the end, as this template one wasn't working

src/main/java/uk/gov/laa/pfla/auth/service/dao/MappingTableDao.java
```diff
@@ -1,53 +1,116 @@
...
+    public List<MappingTableModel> fetchReportList() {
+        ResultSet rslt;
+        int rowNumber = 0;
         mappingTableObjectList.clear(); // Prevent response data accumulating after multiple requests
+        MappingTableModel mappingTableObject;
+
 
-        //fetch data from database
-        //fetchSingleRowOfData()
+        Connection conn = null;
```
deepujames1234, line 87: Can we avoid calling this each time method is called? Jdbc template can be a better solution

Lukehub1: How does JDBC do it then - is it keeping a session open and reusing the same one till you close it or something?

src/main/java/uk/gov/laa/pfla/auth/service/services/MappingTableService.java
```diff
@@ -23,7 +31,8 @@ public MappingTableService(MappingTableDao mappingTableDao) {
         this.mappingTableDao = mappingTableDao;
     }
 
-    public List<ReportListResponse> createReportListResponseList() {
+    public List<ReportListResponse> createReportListResponseList() throws Exception {
+        mappingTableDao.setupDB();
```
deepujames1234, line 35: Can we avoid calling setupDB each time?

deepujames1234 (changes requested):
Setup database is called for each call to getReportList
Why we are not using Jdbctemplate?

src/main/java/uk/gov/laa/pfla/auth/service/services/MappingTableService.java
Lukehub1: That would make sense, yeh i'll look to see how else to arrange the flow

deepujames1234 (approved):

deepujames1234 (approved):

PR #47, opened 2024-11-13
LPF-480: Create new and improved pipelines for payforlegalaid service

SineadCummings:
* Add pipeline that runs on manual trigger and also on creation and edit of PR


H3ll3m4 (approved):
🚀

PR #54, opened 2024-11-21
Feature/lpf 473 cis to ccms import analysis csv

smallmanj-moj:
added test class for MappingTableService.

Other parts of the ticket were addressed by eye:

Data should be consistent with original CCMS invoice analysis report generated by Financial Systems team
Make sure the data is consistent with date the report is generated.
Make sure it has all headers as in CCMS invoice analysis report excel file.


RobertBuczek (dismissed):

RobertBuczek:
It looks really nice just if I may make a suggestion it would be nice to think about:
1) **Magic Numbers**:The test directly uses values like 0, 1000, and 3 in exception-related tests. These could be replaced with constants or descriptive variables:
```
private static final int INVALID_LOW_ID = 0;
private static final int INVALID_HIGH_ID = 1000;
```


2. **Error Message Validation**: Tests for exceptions `should_return_report_id_not_found_exception_when_report_id_out_of_range` only check the type of the exception. It’s a good practice to validate the exception message to ensure that the correct error is being thrown.
```
Throwable exception = assertThrows(ReportIdNotFoundException.class, () -> 
    classUnderTest.getDetailsForSpecificReport(3)
);
assertEquals("Report ID not found: 3", exception.getMessage());
```
3. **Missing scenario**: There’s no test for what happens if `fetchReportList()` returns an empty list. A test like this should be added to validate behavior in such a case:
```
when(mappingTableDao.fetchReportList()).thenReturn(new ArrayList<ReportListEntry>());
```

src/test/java/uk/gov/laa/pfla/auth/service/service/MappingTableServiceTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.laa.pfla.auth.service.dao.MappingTableDao;
+import uk.gov.laa.pfla.auth.service.exceptions.DatabaseReadException;
+import uk.gov.laa.pfla.auth.service.exceptions.ReportIdNotFoundException;
+import uk.gov.laa.pfla.auth.service.models.MappingTableModel;
+import uk.gov.laa.pfla.auth.service.responses.ReportListEntry;
+import uk.gov.laa.pfla.auth.service.services.MappingTableService;
+
+import java.io.IOException;
```
ChrisCookOC, line 15: IOException and ArrayList are unused imports

smallmanj-moj: removed

smallmanj-moj:
Closing as a new PR has been created because of signing issues

PR #57, opened 2024-11-25
[LPF-465]: Generate All Data Models Returned by the Service from the Swagger Documentation [Single point of truth] - Remove builders to data factories

RobertBuczek:
This PR refactors the test data creation logic by removing the builder pattern and introducing a **Test Data Factory**. The goal is to improve the readability and clarity of test data generation, making it easier to understand and work with.

**_Motivation_**:
The existing _*TestBuilder_ used a builder pattern to create ReportsGet200ResponseReportListInner objects. While the builder pattern is useful in some scenarios, in this case, it made the code more difficult to read and maintain. The builder required multiple chained method calls to set default values, and it wasn't immediately clear what the generated data looked like.

In response, I introduced a _Test Data Factory_ that generates the test data in a more straightforward and readable way.


SineadCummings (approved):
Great size!

ChrisCookOC (approved):

PR #58, opened 2024-11-25
[LPF-465]: Generate All Data Models Returned by the Service from the Swagger Documentation [Single point of truth] - Write more unit tests

RobertBuczek:
This PR introduces multiple updates aimed at improving the testing and functionality of the `MappingTableService` and related components. These changes include the creation of test data factories, improvements to the service logic, and the addition of new test cases to ensure comprehensive test coverage.

In MappingTableService.java, the logic of the service was refined, particularly for the methods `fetchReportListEntries` and `getDetailsForSpecificReport`. The `fetchReportListEntries` method was streamlined to fetch and map report entries more efficiently, while the `getDetailsForSpecificReport` method was enhanced with better validation for report IDs, including checks for valid ranges and appropriate error handling. The class utilizes a MappingTableDao to fetch the report data and a mapper to transform the data into the response format expected by the API.

### DIFF
--- a/src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
+++ b/src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
@@ -1,0 +1,197 @@
+package uk.gov.laa.gpfd.service;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import uk.gov.laa.gpfd.dao.MappingTableDao;
+import uk.gov.laa.gpfd.data.MappingTableTestDataFactory;
+import uk.gov.laa.gpfd.exception.DatabaseReadException;
+import uk.gov.laa.gpfd.exception.ReportIdNotFoundException;
+import uk.gov.laa.gpfd.mapper.ReportsGet200ResponseReportListInnerMapper;
+import uk.gov.laa.gpfd.model.MappingTable;
+import uk.gov.laa.gpfd.services.MappingTableService;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class MappingTableServiceTest {
+
+    @Mock
+    private MappingTableDao mappingTableDao;
+
+    @InjectMocks
+    private MappingTableService mappingTableService;
+
+    @BeforeEach
+    void setUp() {
+        MockitoAnnotations.openMocks(this);
+    }
+
+    @Test
+    void shouldReturnReportListEntriesWhenValidReportsExist() {
+        // Given
+        var mappingTable1 = MappingTableTestDataFactory.aValidBankAccountReport();
+        var mappingTable2 = MappingTableTestDataFactory.aValidInvoiceAnalysisReport();
+        var list = Arrays.asList(mappingTable1, mappingTable2);
+        var expected = list.stream().map(ReportsGet200ResponseReportListInnerMapper::map).toList();
+
+        when(mappingTableDao.fetchReportList()).thenReturn(list);
+
+        // When
+        var result = mappingTableService.fetchReportListEntries();
+
+        // Then
+        assertNotNull(result);
+        assertEquals(expected.size(), result.size());
+        assertEquals(expected, result);
+        verify(mappingTableDao, times(1)).fetchReportList();
+    }
+
+    @Test
+    void shouldReturnEmptyListWhenNoReportsExist() {
+        // Given
+        when(mappingTableDao.fetchReportList()).thenReturn(Collections.emptyList());
+
+        // When
+        var result = mappingTableService.fetchReportListEntries();
+
+        // Then
+        assertNotNull(result);
+        assertTrue(result.isEmpty());
+        verify(mappingTableDao, times(1)).fetchReportList();
+    }
+
+    @Test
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
```diff
@@ -0,0 +1,197 @@
...
+        var result = mappingTableService.fetchReportListEntries();
+
+        // Then
+        assertNotNull(result);
+        assertEquals(2, result.size());  // Only one valid mapping should be returned
+        verify(mappingTableDao, times(1)).fetchReportList();
+    }
+
+    @Test
+    void shouldThrowIndexOutOfBoundsExceptionWhenRequestedIdIsLessThanZero() {
```
SineadCummings, line 129: name should be WhenRequestIdIsZero, negative case handled further down

src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
```diff
@@ -0,0 +1,197 @@
...
+        assertNotNull(result);
+        assertEquals(largeList.size(), result.size());
+        verify(mappingTableDao, times(1)).fetchReportList();
+    }
+
+    @Test
+    void shouldReturnValidReportsAndIgnoreInvalidData() throws DatabaseReadException {
+        // Given
+        var mappingTable1 = MappingTableTestDataFactory.aValidInvoiceAnalysisReport();
+        var mappingTable2 = new MappingTable(999, null, null, null, 0, null, null, null, null, null, null); // Invalid data
```
SineadCummings, line 115: Have we confirmed null values are invalid?

src/test/java/uk/gov/laa/gpfd/service/UserServiceTest.java
```diff
@@ -2,68 +2,177 @@
...
+        var exception = assertThrows(AuthUserNotFoundException.class, () -> userService.getUserDetails(mockOAuth2Client));
+
+        // Then
+        assertEquals("No user details returned from Graph", exception.getMessage());
+    }
+
+    @Test
+    void shouldThrowAuthUserNotFoundExceptionWhenGraphUserDataIsIncomplete() {
+        // Given
+        var incompleteUser = new User(); // Missing userPrincipalName and email
```
SineadCummings, line 76: Currently reads as if email is also mandatory - doesn't look to be the case. Can test be updated to use a User that is only missing mandatory fields (userPrincipalName)

SineadCummings (changes requested):
Couple of small items to resolve

src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
RobertBuczek: Thank you for the suggestion! I agree that "WhenRequestIdIsZero" is a clear and descriptive name. However, I opted to go with the negative value scenario here to maintain consistency with the handling logic further down.

src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
RobertBuczek: Thank you for the suggestion! I agree that "WhenRequestIdIsZero" is a clear and descriptive name. However, I opted to go with the negative value scenario here to maintain consistency with the handling logic further down.

src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
RobertBuczek: I wanted to explicitly express the data that we might receive from the database to better align with and accurately map the model.

src/main/java/uk/gov/laa/gpfd/model/MappingTable.java
```diff
@@ -1,22 +1,67 @@
...
-        String csvName,
-        int excelSheetNum,
-        String sqlQuery,
-        String baseUrl,
-        String reportOwner,
-        String reportCreator,
-        String description,
-        String ownerEmail
-) {}
+        Optional<Integer> id,
```
ChrisCookOC, line 10: Is there a reason these all need to be optional? Surely things like an id is rather essential?

src/test/java/uk/gov/laa/gpfd/service/MappingTableServiceTest.java
ChrisCookOC: Test is duplicated below shouldThrowIndexOutOfBoundsExceptionWhenRequestedIdIsNegative

src/main/java/uk/gov/laa/gpfd/model/MappingTable.java
RobertBuczek: Thank you for the question!

The reason I moved to using explicitly `Optional` for all fields, including `id`, is to make explicit the structure of the data as it might be retrieved from the database, especially in cases where each column may be nullable. 

In the case of the `CSV_TO_SQL_MAPPING_TABLE`, table going to be created from: 
```
"GPFD"."CSV_TO_SQL_MAPPING_TABLE" 
   (	
        "ID" NUMBER(*,0), 
	"REPORT_NAME" VARCHAR2(200 BYTE), 
	"SQL_QUERY" VARCHAR2(4000 BYTE),
	"BASE_URL" VARCHAR2(450 BYTE), 
	"REPORT_OWNER" VARCHAR2(100 BYTE), 
	"REPORT_CREATOR" VARCHAR2(100 BYTE), 
	"REPORT_DESCRIPTION" VARCHAR2(400 BYTE), 
	"EXCEL_REPORT" VARCHAR2(200 BYTE), 
	"EXCEL_SHEET_NUM" NUMBER, 
	"CSV_NAME" VARCHAR2(200 BYTE), 
	"OWNER_EMAIL" VARCHAR2(300 BYTE)
   ) 
   ```
where fields like `id`,`reportName`, or `sqlQuery` might be missing or have null values, depending on the specific records we retrieve.

While it's true that in many contexts, an `id` is essential and typically expected, the use of `Optional` allows us to safely handle cases where data might be missing (defensive programming) without introducing `NullPointerExceptions`.

By using `Optional`, we explicitly express that certain fields might be absent and allow the calling code to handle those cases appropriately, making the model more flexible and robust when interacting with the database.

SineadCummings (approved):

ChrisCookOC (approved):

PR #62, opened 2024-11-27
[LPF-465]: Generate All Data Models Returned by the Service from the Swagger Documentation [Single point of truth] - Enable local run

RobertBuczek:
This pull request introduces Liquibase integration and conditional Spring Security bean configuration.

The Liquibase configuration is implemented using a`SpringLiquibase` bean, which is initialized with the writeDataSource and references the changelog file `db.changelog-master.xml`. This enables automatic execution of database changes on application startup.

Additionally, the`SpringLiquibase` bean is conditionally created based on the `spring.liquibase.enabled` is set to`true` in the application's configuration, the`Liquibase` bean will be instantiated and the database migrations will be applied. This conditional setup provides flexibility, allowing Liquibase to be disabled in specific environments, such as during testing or local development.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -271,9 +271,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.liquibase</groupId>
+            <artifactId>liquibase-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.h2database</groupId>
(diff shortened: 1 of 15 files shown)


ChrisCookOC:
Closing this for now. It is Liquibase related and that implementation is unnecessary right now

PR #78, opened 2024-12-16
LPF-540:  Integration tests for /csv/{id} endpoint

SineadCummings:
Add integration tests for /csv/{id} endpoint


src/test/java/uk/gov/laa/gpfd/integration/OAuth2TestConfig.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Bean
+    public ClientRegistrationRepository clientRegistrationRepository() {
+        ClientRegistration graphRegistration = ClientRegistration
+                .withRegistrationId("graph")
+                .clientId("test-client-id")
+                .authorizationUri("authorizationUri")
+                .tokenUri("tokenUri")
+                .redirectUri("redirectUri")
+                .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE) // Specify the authorization grant type
```
ChrisCookOC, line 29: Might be an unnecessary comment?

src/test/java/uk/gov/laa/gpfd/integration/ServerSideErrorIT.java
```diff
@@ -37,4 +55,44 @@ void getReportWithIdShouldReturn500WhenCannotConnectToDb() throws Exception {
 
         Assertions.assertEquals(500, response.getStatus());
     }
+
+    @Test
```
ChrisCookOC, line 59: Thought - if these tests were run in a different order, would this test break? 
So if test below it created a db then this one ran, would it be upset?

SineadCummings: For that to happen we would need to have set up parallel execution. Whilst the tests may be executed in a different order getCsvWithIdShouldReturn500WhenCannotConnectToDbForReportTable both sets up and drops the tables required for testing. As a result, no matter when getCsvWithIdShouldReturn500WhenCannotConnectToDbForMappingTable runs, there will not be a mapping table to point at

ChrisCookOC: oh yeah it drops the tables thats fine :)

ChrisCookOC (approved):

PR #79, opened 2024-12-16
LPF-539: Fix development environment

H3ll3m4:
The development environment was not configured with the new service laa-get-payments-finance-data-dev. We were not using the right spring user profile. We have registered a new application on the dev tenant of Azure Entra ID using the correct url. It does now display the right service laa-get-payments-finance-data-dev and not laa-pay-for-la-dev in the url.

[JIRA](https://dsdmoj.atlassian.net/browse/LPF-539)


ChrisCookOC (approved):

SineadCummings (approved):
looks good!

PR #80, opened 2024-12-16
LPF-532: Add Bearer to swagger definition

SineadCummings:


src/main/resources/static/swagger.yml
```diff
@@ -66,9 +73,14 @@ components:
                 type: string
                 example: "BadRequestException: The request could not be processed."
 
+security:
```
ChrisCookOC, line 76: Does having this here not define it for all the endpoints below anyway?

SineadCummings:
To test this change, the feature branch was deployed to dev to confirm user can still access `/reports` `/reports/1` and `/csv/1`when logged in. Adding security config to swagger acts as documentation only. Authorisation is handled by Azure and OAuth2

image

ChrisCookOC (approved):

PR #81, opened 2024-12-16
LPF-513 - add Snyk to pipelines

ChrisCookOC:
- Add Scan Docker Image pipeline that runs on schedule each day and uploads results to GitHub Security
- Add Snyk scan to Feature branch pipeline checking for high and critical issues
- Add Snyk scan to Deployment pipelines checking for high and critical issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
         <apache-commons-csv.version>1.10.0</apache-commons-csv.version>
 
         <swagger-annotations.version>2.2.10</swagger-annotations.version>
-        <springdoc-oas.version>2.6.0</springdoc-oas.version>
+        <springdoc-oas.version>2.8.3</springdoc-oas.version>
 
         <bytebudy.version>1.15.10</bytebudy.version>
         <mockito-inlin.version>3.11.2</mockito-inlin.version>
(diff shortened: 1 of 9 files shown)


.github/workflows/scan_docker_image.yml
```diff
@@ -0,0 +1,63 @@
+name: Scan docker image
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Run at 7:30AM UTC every day
```
SineadCummings, line 6: Is this standard? Do we need to discuss this setting as a team?

SineadCummings (changes requested):
Need confirmation of inclusion of cron job

.github/workflows/scan_docker_image.yml
ChrisCookOC: Everywhere in LAA that does a container scan does it as a cron job (unless its during their deploy pipeline)
https://github.com/search?q=org%3Aministryofjustice%20github%2Fworkflows%20%2Fsnyk.yml&type=code
https://github.com/search?q=org%3Aministryofjustice+snyk%2Factions%2Fdocker&type=code

The idea being that you are checking the container for changes in vulnerabilities every day, even if you aren't pushing new code, because you want to be aware of any issues with services even if they are live and not getting heavily developed.

As for the time it runs, there is no real science behind that choice.

RobertBuczek (dismissed):

SineadCummings (approved):

RobertBuczek (approved):

PR #90, opened 2025-01-07
Bug/lpf 543 wrong report being returned

smallmanj-moj:
Amended the function to retrieve the report so it returns the right report


src/main/java/uk/gov/laa/gpfd/services/MappingTableService.java
```diff
@@ -69,14 +69,26 @@ public MappingTable getDetailsForSpecificMapping(int requestedId)  {
...
 
-        int indexInt = requestedId - 1;   // The '-1' accounts for the fact that the array index starts at 0, whereas the database index/id starts at 1
-        log.debug("Checking the reportListResponses for the desired reportListResponse object, the requested report index is: {}", indexInt);
+        log.debug("Checking the reportListResponses for the desired reportListResponse object, the requested report ID is: {}", requestedId);
 
-        if (indexInt >= reportListResponses.size()) {
+        return getMappingTable(requestedId, reportListResponses);
+    }
+
+    public MappingTable getMappingTable(int requestedId, List<MappingTable> reportListResponses) {
```
RobertBuczek, line 77: This code currently fetches the entire entity from the database within the method, even when only a single object with a specific requestedId is required.

Fetching the entire list has the following potential drawbacks:

- Increased Memory Usage: Loading the entire list into memory consumes significantly more memory than fetching only the required object. This can be particularly problematic for large datasets, potentially leading to **OutOfMemoryErrors** or impacting application performance.
- Unnecessary Network Traffic: Fetching the entire list involves transferring more data over the network than necessary, increasing network latency and potentially impacting overall system performance, especially in distributed environments.
- Unnecessary Processing: Iterating through the entire list to find the desired object consumes CPU cycles that could be saved by directly fetching the specific object.

smallmanj-moj: function incorporated into calling function.

H3ll3m4: Yes I agree with @RobertBuczek it would be great to optimise this request while it is still fresh in our minds

ChrisCookOC: Yeah agreed we don't want to get the whole database, which does involve changing how we do that endpoint. Do we want to do that on this ticket or a new ticket?

smallmanj-moj: functionality has been changed to add new methods which separate getting all reports from getting one report

src/main/java/uk/gov/laa/gpfd/dao/MappingTableDao.java
```diff
@@ -19,18 +19,19 @@
 @Repository
 @RequiredArgsConstructor
 public class MappingTableDao {
-    private static final String SELECT_SQL = "SELECT * FROM GPFD.CSV_TO_SQL_MAPPING_TABLE";
+    private static final String SELECT_SQL = "SELECT * FROM GPFD.CSV_TO_SQL_MAPPING_TABLE ORDER BY ID ASC";
+    private static final String SELECT_SINGLE_ITEM_SQL = "SELECT * FROM GPFD.CSV_TO_SQL_MAPPING_TABLE WHERE ID = ";
```
ChrisCookOC, line 23: I'd be tempted to make it a function that you pass the requestId in rather than append it to the constant in the code. but this is fine also.

src/test/java/uk/gov/laa/gpfd/dao/ReportViewsDaoTest.java
```diff
@@ -0,0 +1,59 @@
...
+        String sqlData = FileUtils.readResourceToString("gpfd_data.sql");
+
+        writeJdbcTemplate.execute(sqlSchema);
+        writeJdbcTemplate.execute(sqlData);
+    }
+
+    @AfterEach
+    void resetDatabase() {
+
+        writeJdbcTemplate.update("TRUNCATE TABLE GPFD.REPORT_TRACKING");
```
ChrisCookOC, line 42: This and the BeforeEach are the same in each three DAO test. Wonder if we want to make them helper utils?

ChrisCookOC (commented):
Added a comment that is more just preference than requirements, so feel free to ignore.

But I will suggest you check out the three issues SonarCloud has raised in it's comment above

src/main/java/uk/gov/laa/gpfd/dao/MappingTableDao.java
```diff
@@ -47,6 +48,19 @@ public List<MappingTable> fetchReportList() throws DatabaseReadException {
         } catch (MappingException e) {
             log.error("Exception with model map loop: %s", e);
         }
+    }
+
+    public MappingTable fetchReport(int requestId) throws DatabaseReadException {
+        List<MappingTable> mappingTableObjectList = new ArrayList<>();
+        String sqlCommand = SELECT_SINGLE_ITEM_SQL + requestId;
```
RobertBuczek, line 55: This code is vulnerable to SQL injection. The requestId is directly concatenated into the SQL query string, allowing malicious input to be executed as SQL commands.

Recommendations:
- Replace the current string concatenation with a prepared statement.
- Bind the requestId as a parameter to the prepared statement.
- Validate the requestId 

Docs: https://docs.oracle.com/javase/8/docs/api/java/sql/PreparedStatement.html

src/main/java/uk/gov/laa/gpfd/services/MappingTableService.java
smallmanj-moj: functionality changed to include parameters in queries. No string concatenation now.

src/test/java/uk/gov/laa/gpfd/dao/ReportViewsDaoTest.java
smallmanj-moj: I think this should be logged as tech debt.

src/main/java/uk/gov/laa/gpfd/dao/MappingTableDao.java
smallmanj-moj: This has been changed to use parameters

ChrisCookOC (approved):

RobertBuczek (approved):

PR #93, opened 2025-01-09
LPF-557 - fix CSV export on local

ChrisCookOC:
Resolves issue where CSV export endpoint would crash when running on a profile that did not have azure auth turned on, such as Local

Also adds some test data so that the h2 database on local is populated - without data it will return a 500.


RobertBuczek:
I'm very impressed with the progress made so far. To that end, I'd like to offer some feedback on this approach. I believe these suggestions may help us further strengthen the plan and ensure its long-term success:

- **Production Code Integrity**: Adding test-specific data and structures directly into production code is risky. It increases the likelihood of production issues, especially if the test data is accidentally left in place or executed in a live environment. Production schemas and tables should represent only the structures and data essential for the live system's functionality.

- **Separation of Concerns:** Acceptance testing and production environments serve distinct purposes. A better practice would be to set up a separate schema, table, or database specifically for testing. This keeps the test-related resources isolated from production.

- **Data Validity and Types**: The INSERT statement introduces test data that doesn't align with realistic production scenarios (e.g., large and inconsistent values like '213239999494' and '7'). This could mislead future developers or testers who assume these values are representative.

**Suggestion**: 
Instead of modifying production code, you could create a dedicated schema or use a database mocking tool for acceptance tests. Many testing frameworks allow seeding a test database with data during setup without impacting production code. I'd be happy to help you explore these options if you'd like!

Thanks again for your efforts here. Let me know if you’d like to discuss this further!

smallmanj-moj (dismissed):

ChrisCookOC:
Switched to use h2 in file mode so the test can add its own data.

RobertBuczek (approved):

PR #110, opened 2025-01-15
Feature/lpf 551 create UUID for primary keys

RobertBuczek:
Update DDLS


smallmanj-moj (changes requested):
Why do we regenerate the UUID? If we use constant value UUID then it will simplify maintenance should we re run the scripts. Look at what Deepu said about putting the ID in a reports table for post MVP. This approach would mean that table will need regenerating.

smallmanj-moj (approved):

PR #142, opened 2025-02-04
LPF-602 - change network policy to allow Prometheus through

ChrisCookOC:
Change Network Policy podselector to allow the monitoring namespace (prometheus) to access **any** pod in our namespace. Before it was only looking for our service pod, and so could not access our actuator endpoint.

This setup is used in network policies in many MoJ repos. 

We can see this works by comparing dev (where it is deployed) and uat (where it is not) - you can tell if Prometheus can access the metrics by looking at the targets section of Prometheus
image

We can also test the metrics that were failing previously on the ticket, e.g. jvm_info
Works on Dev now
image

Doesn't on uat still
image

### DIFF
--- a/.github/workflows/deploy_to_dev.yml
+++ b/.github/workflows/deploy_to_dev.yml
@@ -2,7 +2,6 @@ on:
   push:
     branches:
       - main
-      - feature/LPF-602-network
 
 jobs:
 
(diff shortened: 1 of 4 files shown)


H3ll3m4 (approved):
🚀

RobertBuczek (approved):

PR #151, opened 2025-02-25
Addition of dao classes and new database structure

smallmanj-moj:
addition of classes, database structures and reference data


RobertBuczek:
No longer required

PR #158, opened 2025-02-28
LPF-570 - fix docker scan

ChrisCookOC:
Docker scan fails due to the changes around the open api. example failure: https://github.com/ministryofjustice/payforlegalaid/actions/runs/13625682748/job/38082398692
image

Need to supply the settings.xml like we have on other pipelines to fix this
with fix: https://github.com/ministryofjustice/payforlegalaid/actions/runs/13587413293 
image


H3ll3m4:
Can we add a short comment in the description, what was the issue, as well as evidence, for example screenshot of a successful deployment?

H3ll3m4 (approved):
🚀

PR #171, opened 2025-03-12
[LPF-626]: Create CCMS Invoice Analysis report

RobertBuczek:
It will fail until Spec is not redy


src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -21,18 +21,26 @@
...
+import java.util.Optional;
 import java.util.UUID;
 
 @Slf4j
 @RestController
 @RequiredArgsConstructor
-public class ReportsController implements ReportsApi {
+public class ReportsController implements ReportsApi, ExcelApi {
+
+    private static final String foo = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet";
```
SineadCummings, line 32: do we want to use foo.....

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -21,18 +21,26 @@
...
 
     private final ReportTrackingTableService reportTrackingTableService;
     private final ReportService reportService;
     private final ReportManagementService reportManagementService;
     private final StreamingService streamingService;
 
+    @Override
+    public Optional<NativeWebRequest> getRequest() {
+        return Optional.empty();
```
SineadCummings, line 41: Need a bit of explanation for this one please!

RobertBuczek: When a class implements multiple interfaces that have methods with the same signature (i.e., the same method name and parameter list), it can lead to ambiguity. In our case, both `ExcelApi` and `ReportsApi` interfaces have a default method `getRequest()` with the same signature. This creates a situation where the compiler cannot determine which `getRequest()` method should be used when the method is called on an instance of `ReportsController`.

SineadCummings (dismissed):

H3ll3m4 (approved):
🚀

SineadCummings (approved):

H3ll3m4 (approved):
🚀

PR #178, opened 2025-03-21
LPF-707: Add reports tracking functionality 

vvartakMoJ:
The [JIRA issue](https://dsdmoj.atlassian.net/browse/LPF-707) talks about a report creation and a report download as two separate possibilities but currently we do not have the download (from secure storage cache) option, so the user is always recorded as the "report creator" in the reports_tracking table. We will have to deal with tracking of the downloads when the functionality is built.

I have also refactored the tests (both integration and unit tests) by:

1. extracting the code for setting up and clearing of the test data into a utils class (DatabaseUtils) and calling it from base classes.
2. Replace the SQL files with equivalent Liquibase XML files to reduce platform dependent code.


src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -66,11 +63,9 @@ public ResponseEntity<ReportsGet200Response> reportsGet() {
      * @return CSV data stream or reports data
      */
     @RequestMapping(value = "/csv/{id}", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
-    public ResponseEntity<StreamingResponseBody> getCSV(@PathVariable(value = "id") UUID requestedId,
-                                                        @RegisteredOAuth2AuthorizedClient("graph") OAuth2AuthorizedClient graphClient) {
-        reportTrackingTableService.updateReportTrackingTable(requestedId, graphClient);
-
+    public ResponseEntity<StreamingResponseBody> getCSV(@PathVariable(value = "id") UUID requestedId) {
         log.debug("Returning a CSV response to user");
+        reportsTrackingService.saveReportsTracking(requestedId);
```
RobertBuczek, line 68: I was wondering if it might be a good idea to move Tracking information publishing off the main thread to a background process. Since this operation doesn’t need to block the main thread, doing so could improve performance without impacting functionality. Would love to hear your thoughts!

src/main/java/uk/gov/laa/gpfd/services/UserService.java
```diff
@@ -52,4 +55,20 @@ public UserDetails getUserDetails(OAuth2AuthorizedClient client) {
...
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        String username = "Unknown";
+        if (authentication  != null && authentication.getPrincipal() != null) {
+            Object principal = authentication.getPrincipal();
+
+            if (principal instanceof DefaultOidcUser) {
+                username = ((DefaultOidcUser) principal).getAttribute("preferred_username");
+            } else {
+                username = principal.toString()
+                    .substring(0, 100); // Fallback in case it's a simple string
```
H3ll3m4, line 68: Should that be the lenght of the principal.toString()?

src/test/java/uk/gov/laa/gpfd/service/ReportsTrackingServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+import static org.mockito.Mockito.when;
+
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
```
H3ll3m4, line 19: Is OAuth2AuthorizedClient still used?

src/it/java/uk/gov/laa/gpfd/integration/BaseIT.java
```diff
@@ -0,0 +1,46 @@
...
+    databaseUtils.setUpDatabase();
+  }
+
+  @AfterAll
+  void cleanUpDatabase() {
+    databaseUtils.cleanUpDatabase();
+  }
+
+  @NotNull
+  protected MockHttpServletResponse getResponseForAuthenticatedRequest(String uriTemplate) throws Exception {
```
RobertBuczek, line 37: While the current helper method `getResponseForAuthenticatedRequest` provides a convenient way to mock authenticated requests, there are a few potential downsides worth considering:
- Maintenance Risks: If the authentication logic evolves (e.g., new claims or client requirements), all tests depending on this method would need updates, creating a tight coupling.
- Tests using this helper may hide important authentication/authorization details behind a generic call, making it harder to understand the specific security requirements being verified.

src/test/java/uk/gov/laa/gpfd/dao/ReportsTrackingDaoTest.java
```diff
@@ -0,0 +1,67 @@
...
+        Timestamp reportCreationTime = (Timestamp) reportsTrackingList.get(1).get("CREATION_DATE"); //index 1 because index 0 is populated by liquibase
+
+        assertEquals(2, reportsTrackingList.size());
+        Assertions.assertNotNull(reportsTrackingList.get(1).get("ID"));
+        assertEquals(insertedReportName, reportName);
+        assertEquals(insertedReportUrl, reportsTrackingList.get(1).get("REPORT_URL"));
+        assertEquals(insertedCreationTime, reportCreationTime);
+        assertEquals(insertedTemplateUrl, reportsTrackingList.get(1).get("TEMPLATE_URL"));
+    }
+}
```
H3ll3m4, line 67: Do we need to test invalid data? Null or duplicate entries?

src/it/java/uk/gov/laa/gpfd/integration/GetReportsIT.java
```diff
@@ -67,7 +39,7 @@ void shouldReturnListOfReports() throws Exception {
 
         var json = new JSONObject(response.getContentAsString());
         JSONArray reportList = (JSONArray) json.get("reportList");
-        Assertions.assertEquals(3, reportList.toList().size());
+        Assertions.assertEquals(4, reportList.toList().size());
```
RobertBuczek, line 42: The assertion `Assertions.assertEquals(4, reportList.toList().size())` uses a "magic number" (4), which might make the test less maintainable or harder to understand at a glance.

src/test/java/uk/gov/laa/gpfd/dao/ReportsTrackingDaoTest.java
vvartakMoJ: Added tests for invalid/null data. Duplicate entries are not possible to test (the PK ID is auto-generated, so cannot be duplicated and there is no other unique key).

src/main/java/uk/gov/laa/gpfd/dao/ReportsTrackingDao.java
```diff
@@ -0,0 +1,34 @@
...
+@RequiredArgsConstructor
+public class ReportsTrackingDao {
+  private static final String INSERT_SQL = "INSERT INTO GPFD.REPORTS_TRACKING"
+      + " (ID, NAME, REPORT_ID, CREATION_DATE, REPORT_DOWNLOADED_BY, REPORT_CREATOR, REPORT_OWNER, REPORT_OUTPUT_TYPE, TEMPLATE_URL, REPORT_URL)"
+      + " VALUES (?,?,?,?,?,?,?,?,?,?)";
+  private static final String SELECT_SQL = "SELECT * from GPFD.REPORTS_TRACKING";
+
+  private final JdbcTemplate writeJdbcTemplate;
+
+  public void saveReportsTracking(ReportsTracking reportsTracking) {
```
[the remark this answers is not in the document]
vvartakMoJ: To ensure that a failed tracking call doesn't affect the user, I have added try-catch around it so that it only logs an error and carries on.

src/main/java/uk/gov/laa/gpfd/dao/ReportsTrackingDao.java
```diff
@@ -0,0 +1,34 @@
...
+    log.info("Saving tracking information");
+    int numberOfRowsAffected = this.writeJdbcTemplate.update(INSERT_SQL, UUID.randomUUID().toString(), reportsTracking.getName(), reportsTracking.getReportId(),
```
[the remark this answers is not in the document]
vvartakMoJ: Yes, good comment. I have added try-catch to log an error if this fails. It allows the report generation to carry on if the report tracking fails for any reason.

src/it/java/uk/gov/laa/gpfd/integration/GetReportsIT.java
vvartakMoJ: Magic has been removed!

src/main/java/uk/gov/laa/gpfd/services/UserService.java
vvartakMoJ: The output of principal.toString can be very long (it's a json document with a lot of information). The output needs to be stored in the REPORT_CREATOR column which is defined as follows in liquibase, hence the limit of 100:

<column name="REPORT_CREATOR" type="VARCHAR(100)">

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
vvartakMoJ: I've added a separate thread, please check and see if it's enough.

H3ll3m4 (dismissed):

RobertBuczek (approved):

H3ll3m4 (approved):

PR #179, opened 2025-03-25
[LPF-599] Adding deployment templates to DRY our pipeline

H3ll3m4:
This PR is to DRY our pipeline. We are using reusable deployment workflow deploy to env for DEV, UAT & PROD. This way we are using the same templates yaml files for all the resources and we can delete those files and folders development, uat and prod. We are using one deployment pipeline and thanks to the deployment protection rules, there is still the manual trigger for UAT and PROD : 
image

Deploys our pod properly:
image

In the future, we will update this pipeline so the ATs run successfully before deploying to the higher environments with the rule needs.


SineadCummings (dismissed):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,67 @@
+name: One to deploy them all
+
+on:
+  push:
+    branches:
+      - main
+      - deployment-template
```
vvartakMoJ: Just a question: Does this branch (deployment-template) have to stay in this list when we merge into main and the branch is deleted?

SineadCummings (approved):

.github/workflows/deploy.yml
vvartakMoJ: I see that you removed this in the latest commit already, so resolving this conversation.

H3ll3m4: Thanks Vid! I think I made the change when you were reviewing it ah ah

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,67 @@
...
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  To_Dev:
+    name: Deploy to dev
+    uses: ./.github/workflows/deploy_to_env.yml
+    with:
+      github_environment: development
```
vvartakMoJ, line 14: Is it possible to only use github_environment variable and get rid of the other called environment_name? They have the same values in uat and prod, only differ in dev/development.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,67 @@
...
+      alert_severity: laa-get-legal-aid-data-dev
+      secret_name: gpfd-test-secret-01
+    secrets:
+      openapi_package_user: ${{ secrets.OPENAPI_PACKAGE_USER }}
+      openapi_package_password: ${{ secrets.OPENAPI_PACKAGE_PASSWORD }}
+      snyk_token: ${{ secrets.SNYK_TOKEN }}
+      ecr_role_to_assume: ${{ secrets.DEV_ECR_ROLE_TO_ASSUME }}
+      kube_namespace: ${{ secrets.KUBE_NAMESPACE_DEV }}
+      kube_cluster: ${{ secrets.KUBE_CLUSTER_DEV }}
+      kube_token: ${{ secrets.KUBE_TOKEN_DEV }}
```
vvartakMoJ, line 26: Is it possible to remove the _DEV from these secret names (and _UAT, _PROD on later lines) as they are already in separate environments and so in their own namespaces?

vvartakMoJ (commented):
It's looking great already but I've asked a few questions/suggestions, please see if they are valid and easy-ish to implement now. If not, feel free to resolve the conversations and I'll approve immediately.

.github/workflows/deploy.yml
H3ll3m4: Good suggestion Vid! It would require some PRs for the cloud platform environment, where we have added the use of prefix because we were not deploying in the right ECR if I remember well: https://github.com/ministryofjustice/cloud-platform-environments/blob/55f2171d68059b8946c99a5e2a54a3862205fd5b/namespaces/live.cloud-platform.service.justice.gov.uk/laa-get-payments-finance-data-dev/resources/ecr.tf#L33
And probably a bit more work that I don't know about. I think that's a good idea but harder than the previous suggestion.

vvartakMoJ: We have decided that this is more work than the benefits it achieves, so have decided to leave it as it is for now.

vvartakMoJ (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,67 @@
...
+      kube_cert: ${{ secrets.KUBE_CERT_DEV }}
+
+  To_Uat:
+    name: Deploy to UAT
+    needs: To_Dev
+    uses: ./.github/workflows/deploy_to_env.yml
+    with:
+      github_environment: uat
```
RobertBuczek, line 34: We can have 1 property for env which then can do interpolation:
```
${{ secrets[format('{0}_ECR_ROLE_TO_ASSUME', matrix.environment)] }}
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,67 @@
...
+      openapi_package_user: ${{ secrets.OPENAPI_PACKAGE_USER }}
+      openapi_package_password: ${{ secrets.OPENAPI_PACKAGE_PASSWORD }}
+      snyk_token: ${{ secrets.SNYK_TOKEN }}
+      ecr_role_to_assume: ${{ secrets.DEV_ECR_ROLE_TO_ASSUME }}
+      kube_namespace: ${{ secrets.KUBE_NAMESPACE_DEV }}
+      kube_cluster: ${{ secrets.KUBE_CLUSTER_DEV }}
+      kube_token: ${{ secrets.KUBE_TOKEN_DEV }}
+      kube_cert: ${{ secrets.KUBE_CERT_DEV }}
+
+  To_Uat:
```
RobertBuczek, line 29: Do we need a rollback if something go wrong?

.github/workflows/deploy.yml
H3ll3m4: Good suggestion Rob, I will see how I can make it work with this workflow

.github/workflows/deploy.yml
H3ll3m4: Good idea Rob, I was thinking to add conditions once we get the ATs incorporated to it but we could add a rollback to undo the deployment of kubernetes resources. What would you be testing against?

.github/workflows/deploy.yml
H3ll3m4: I think that would be a good improvement but currently our dev environment is a bit messy, like the secrets name has test, instead of dev in the name so it would require a PR to cloud platform environments

PR #182, opened 2025-03-27
lpf-735: add logic to use default auth behaviour when no auth token

SineadCummings:
Subtask of LPF-664
First stage of work to implement JWT validation to allow authentication of users accessing API's via the UI

- doFilter method is not yet implemented in application, requires bean to do so, this will be added in the final PR for LPF-664


ChrisCookOC (changes requested):
As discussed re: doFilter

ChrisCookOC (approved):

ChrisCookOC (approved):

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -0,0 +1,24 @@
...
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+
+@Slf4j
+public class JwtAuthenticationFilter {
```
vvartakMoJ, line 12: Should this class implement Filter (and annotated as @Component) before it takes effect? Can this be done now and tested (or does it depend on the bits happening later)?

vvartakMoJ (approved):
I have asked a question which may or may not be relevant so happy to approve.

PR #183, opened 2025-03-27
[LPF-599] Fix deployment pipeline for production

H3ll3m4:
We also have updated the secrets in AWS Secret Manager and harmonise the variables across environments.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - fix/deployment-prod
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - fix/deployment-prod
```
vvartakMoJ, line 7: Is this staying or should be removed before merging?

vvartakMoJ (approved):
LGTM, apart from one question about the branch in deploy.yml

PR #185, opened 2025-03-28
lpf-737: validate JWT contents

SineadCummings:
Subtask of LPF-664
adds method to validate jwt payload
includes adding app config for clientId and tenantId. 

By adding these to app config, it is now a requirement that these env vars have a value when the service runs. Application files have been updated for local and test to ensure dummy values are available. It is fine that these are dummy values as auth does not require specific values for local and tests.


src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -17,17 +18,76 @@
...
-            // JWT validation logic to come in later PR
+        if (token == null || token.isEmpty()) {
+            filterChain.doFilter(servletRequest, servletResponse);
+        } else {
+            validateJwt(token);
+            filterChain.doFilter(servletRequest, servletResponse);
         }
+    }
+
+    public boolean validateJwt(String token) {
```
RobertBuczek, line 46: The current `validateJwt()` method returns a boolean, but it always throws exceptions on failure rather than returning `false`. This can be misleading since callers might expect a `false` return for invalid tokens

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -17,17 +18,76 @@
 public class JwtAuthenticationFilter extends OncePerRequestFilter {
     private static final String TOKEN_PREFIX = "bearer ";
     private static final int TOKEN_PARTS = 3;
+    private static final String JWT_PAYLOAD_TENANT_ID_KEY = "tid";
+    private static final String JWT_PAYLOAD_APPLICATION_ID_KEY = "appid";
+    private static final String SCOPE_KEY = "scp";
+    private static final String SCOPE_VALUE = "User.Read";
```
RobertBuczek, line 24: SCOPE_VALUE is currently hardcoded to `"User.Read"`. While this works for the current use case, it might limit flexibility if:
- Different endpoints require different scopes
- Future requirements need to validate multiple scopes

Suggested Improvement:
Make the required scopes configurable by:
- Accepting a Set<String> parameter in the method, or
- Reuse from service configuration:
```
          graph:
            scopes:
              - https://graph.microsoft.com/User.Read
```

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -17,17 +18,76 @@
...
-        filterChain.doFilter(servletRequest, servletResponse);
+            if (!decodedToken.getClaimAsString(JWT_PAYLOAD_TENANT_ID_KEY).equals(appConfig.getEntraIdTenantId())) {
+                throw new JwtException("Incorrect Tenant ID");
+            }
+
+            if (!decodedToken.getClaimAsString(JWT_PAYLOAD_APPLICATION_ID_KEY).equals(appConfig.getEntraIdClientId())) {
+                throw new JwtException("Incorrect Application ID");
+            }
+
+            if (!isTokenValidForCurrentTime(decodedToken)) {
```
RobertBuczek, line 72: Currently, we check isTokenValidForCurrentTime() before isTokenExpired(). Since expired tokens should always be rejected immediately, we could potentially:
- Check expiry first - This would fail fast for the most common invalid case
- Then validate not-before time
- Finally verify other claims (audience, tenant ID, etc.)

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -17,17 +18,76 @@
...
+                throw new JwtException("Token not valid for current time");
+            }
+
+            if (isTokenExpired(decodedToken)) {
+                throw new JwtException("Token is expired");
+            }
+
+            if (!decodedToken.getClaimAsStringList(SCOPE_KEY).contains(SCOPE_VALUE)) {
```
RobertBuczek, line 80: Currently, when exceptions occur during validation, we wrap them with a `new JwtException("Unable to validate token")`. While this simplifies the error message, it unfortunately loses some valuable debugging context.

For example:
If getClaimAsStringList(SCOPE_KEY) throws a NullPointerException

We'll only see "Unable to validate token" in logs
The original exception (and its stack trace) gets hidden

**Suggested Improvement:**
We could preserve the original exception while still providing a clean error message by:
- Including the original exception as the cause
- Adding its message to our error description

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -6,28 +6,92 @@
...
+    }
 
     @Override
     public void doFilterInternal(HttpServletRequest servletRequest, @NotNull HttpServletResponse servletResponse, @NotNull FilterChain filterChain) throws IOException, ServletException {
         var token = servletRequest.getHeader("Authorization");
 
-        if (token != null && !token.isEmpty()) {
-            // JWT validation logic to come in later PR
+        if (token == null || token.isEmpty()) {
+            filterChain.doFilter(servletRequest, servletResponse);
```
vvartakMoJ, line 42: As this line is duplicated in both if and else, is it better to take it out (in which case the previous if condition might work better).

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -6,28 +6,92 @@
...
+    }
+
+    public boolean validateJwt(String token) {
+        var jwtContent = extractJwtToken(token);
+
+        try {
+            Jwt decodedToken = jwtDecoder.decode(jwtContent);
+
+            if (decodedToken == null)
+                throw new JwtException("decode token returned null");
```
vvartakMoJ, line 56: It might be useful (when reading the logs in order to investigate issues) to add the jwtContent to the message (unless it is a security risk, in which case, perhaps a portion of it might be logged).

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
SineadCummings: As discussed this causes problems in other areas as the scopes value is used for auth and https://graph.microsoft.com/User.Read is expected to be in a single var.

Opting to leave this unaddressed

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
SineadCummings: Resolved as requested

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
SineadCummings: Resolved as requested, now including the message for the generic Exception catch to the throw of new JwtException

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -17,17 +18,76 @@
 public class JwtAuthenticationFilter extends OncePerRequestFilter {
     private static final String TOKEN_PREFIX = "bearer ";
```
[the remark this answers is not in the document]
SineadCummings: Added token component enum

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
SineadCummings: we'll be adding a hashed token to the ogs to improve traceability in the next PR. This is to help identify which errors come from which token. Using hashed token intended to balance traceability with security of token

src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
SineadCummings: void return type reduces testability of method
returning false would require return of error message, throwing exception fails fast
happy to leave as is, may return false in future iterations

vvartakMoJ (approved):
LGTM

RobertBuczek (approved):

PR #186, opened 2025-03-31
Feature/lpf 737a

SineadCummings:
Subtask of LPF-664
adds method to validate timestamps from JWT payload


src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -55,4 +56,19 @@ public String extractJwtToken(String token) {
         return token;
     }
 
-}
+    public boolean isTokenValidForCurrentTime(Jwt decodedToken) {
```
vvartakMoJ, line 59: Would combining these two methods into a single one "isTokenValid" keep things more straight-forward in the calling code?

Also, why not return false in case the thing being checked is null?

All different conditions can be checked in this one place, errors logged to explain exactly what was wrong, but the client should not have to worry about handling all these possibilities (as all it needs to know is whether to proceed or not).

Just personal opinion, so happy to approve if you don't agree.

vvartakMoJ (approved):
LGTM

RobertBuczek (approved):

PR #196, opened 2025-04-14
LPF-740: expose jwt decode logic

SineadCummings:
permits access to payforlegal aid with valid jwt. This is to enable users to access data via the ui
Populates security context


src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -91,12 +104,15 @@ private void validateJwt(String token, String logIdentifier) {
 
             log.info("Token " + logIdentifier + " - JWT validated successfully");
 
+            SecurityContextHolder.getContext().setAuthentication(new UsernamePasswordAuthenticationToken(username, Optional.empty(), List.of()));
```
SineadCummings, line 107: _Comment left by @RobertBuczek on previous PR, moved to this PR_

I really appreciate the careful JWT validation being performed here. I just wanted to gently highlight a few observations about the current authentication setup:

No credentials (Optional.empty())
No granted authorities (List.of())
Which means:

The JWT already includes valuable scope information, but this isn’t currently being mapped to Spring Security authorities.
The JWT validation is wonderfully thorough, but the resulting Authentication object doesn’t fully reflect this e.g., the empty credentials.

SineadCummings:
This work was paused. Originally intended for integration of a UI layer, this UI has now been developed in thymeleaf, within this repository, so no JWT validation layer is required

PR #197, opened 2025-04-14
lpf-740: security context

SineadCummings:
Add security context logic to JWT authentication journey


src/main/java/uk/gov/laa/gpfd/config/JwtAuthenticationFilter.java
```diff
@@ -91,12 +104,15 @@ private void validateJwt(String token, String logIdentifier) {
 
             log.info("Token " + logIdentifier + " - JWT validated successfully");
 
+            SecurityContextHolder.getContext().setAuthentication(new UsernamePasswordAuthenticationToken(username, Optional.empty(), List.of()));
```
RobertBuczek, line 107: I really appreciate the careful JWT validation being performed here. I just wanted to gently highlight a few observations about the current authentication setup:
- No credentials (Optional.empty())
- No granted authorities (List.of())

Which means:
- The JWT already includes valuable scope information, but this isn’t currently being mapped to Spring Security authorities. 
- The JWT validation is wonderfully thorough, but the resulting Authentication object doesn’t fully reflect this e.g., the empty credentials.

PR #211, opened 2025-05-08
Get template from SDS mock

SineadCummings:
* Get template using id, from SDS mock service
* gpfd will call SDS mock service (eventually replaced by real secure document service), get url for required template and then call url to get temaplte as input stream
* SDS endpoint is currently `/retrieve_file` but will soon become `/get_file` - we are using new endpoint name for future proofing
* `/get_file` returns presigned URL from AWS with default expiry as string
*


SineadCummings:
This work was replcaed by use of S3 bucket

PR #219, opened 2025-05-13
LPF-629 - mitigate SQL injection potential 

ChrisCookOC:
We run SQL queries we get from the database directly. This introduces a small risk that the value we get from the database is damaged, malicious etc and could be executed. This applies to both the CSV_TO_SQL_MAPPING_TABLE and the REPORT_QUERY tables.

As an example, lets say the SQL query is now `SELECT * FROM ANY_REPORT.V_MONEY_STUFF; DROP TABLE ANY_REPORT.V_MONEY_STUFF`. Now when the CSV or Excel run it deletes that table. Oops :( 

We check it is a valid sql select using regex 
If we are asked to execute SQL that does not match that format, we get an error
image

image


H3ll3m4 (approved):
LGTM

vvartakMoJ (commented):
Looks good to me, except that now we depend on the query being a very basic one of the very limited format of "SELECT * FROM xxxx" where only xxxx can change, i.e. we don't really need a query from the database, just the name of the view.

Any future developer will think "why didn't they just store the name of the view in the database"?

Could we consider the `jsqlparser` library, which allows us to check if a query is a valid `select` query:

```
Statement stmt = CCJSqlParserUtil.parse(sql);
        return stmt instanceof Select;
```

ChrisCookOC:
Thanks @vvartakMoJ , will take a look at the parser now :) 

Re: tying into the view format. Did consider that, and whether it would be beneficial to in a future change to just store the view - didn't want to just dive in as would involve many changes

But if we are in the future gonna have more sources that might not follow the same format as Mojfin, then using the "is it a select statement" parser check will hopefully be future proof

src/main/java/uk/gov/laa/gpfd/services/excel/ExcelCreationService.java
```diff
@@ -87,6 +90,11 @@ private void updateTemplateWithData(Workbook workbook, Report report) {
      * Fetches data from the database asynchronously for a given SheetToQuery pair.
      */
     private CompletableFuture<List<Map<String, Object>>> fetchData(SheetToQuery sheetToQuery) {
+        if (!sqlFormatValidator.isSqlFormatValid(sheetToQuery.reportQuery().getQuery())) {
```
RobertBuczek, line 93: I was wondering if we could explore the possibility of enhancing our data model class with an annotation or a more robust type definition to ensure validation at the object-creation level.

For example, we currently have:
```
@Immutable  
public abstract class ReportQuery {  
    @Nullable  
    public abstract UUID getId();  
    @Nullable  
    public abstract UUID getReportId();  
    @Nullable  
    public abstract String getQuery();  
    @Nullable  
    public abstract String getTabName();  
    @Nullable  
    public abstract Collection<FieldAttributes> getFieldAttributes();  
}  
```
Specifically, we have:
```
@Nullable  
public abstract String getQuery();  
```
Would it be possible to either:

Replace String with a more structured type (e.g., Object Query) that enforces validity, or

Introduce an annotation that automatically validates whether the query is well-formed?

The goal would be to prevent object construction entirely if the query is invalid—since there’s no benefit in creating a report with invalid data, as it would only consume resources without providing meaningful value.

vvartakMoJ (dismissed):
LGTM!

src/main/java/uk/gov/laa/gpfd/exception/DatabaseReadException.java
```diff
@@ -1,10 +1,22 @@
 package uk.gov.laa.gpfd.exception;
 
 /**
- * Exception class to indicate that an error occurred while reading from the database
+ * A sealed abstract class representing a hierarchy of exceptions related to database read operations.
```
ChrisCookOC, line 4: Will actually make it sealed abstract in a separate PR I've already prepared. Wanted to split it as that involves lots of changes to disparate files and this PR should focus on the sql stuff.

RobertBuczek (approved):

PR #276, opened 2025-07-24
Bugfix: LPF-970 - fix pivot tables needing repairs when opening the downloaded sheet

ChrisCookOC:
JIRA: [LPF-970](https://dsdmoj.atlassian.net/browse/LPF-970)

## Description of changes
- Some reports would show an alert saying they need repairs
image

- This was caused by certain pivot tables - the definitions had reference to pivot table styles (dxf styles) that were not copied into the new report.
- This change now copies all those dxf styles over. It does them in one go since they are only used by analytics sheets and there are a small number. By copying all at once we also ensure that the references in the definition match the references in the style sheet.
- Also, a small bug fix where by the pivot CT cache id was getting overwritten by the cache id - this caused no visible issue (due to how the cached data isn't really needed by us), but by fixing it we no longer generate XML that points at a non-existent cache file.
- Also, fix formatting on the first pivot table. The date format now matches the original report
- Also, ensure pound values are formatted as such on the second report
- Also, ensure report tabs match recent changes made in the data repo

## Evidence
- Before
Download CCMS invoice analysis report - get alert, need Excel to repair the file
https://github.com/user-attachments/assets/6d72634c-40a9-47df-8e41-dd4efd0dba2e

- After:
Download CCMS invoice analysis report - no alert
https://github.com/user-attachments/assets/f262233b-89b7-49d2-971a-d783d992d943

(note: sheet ordering difference due to my data not being setup properly locally)

Formatting in pivots is correct types
image
image

Formulas refer to the new tab names set up in LPF-986 and are still correctly calculated
image

## Checklist
- [ ] Clean commit history

[LPF-970]: https://dsdmoj.atlassian.net/browse/LPF-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/uk/gov/laa/gpfd/services/excel/copier/PivotStyleCopier.java
```diff
@@ -0,0 +1,72 @@
...
+    /** The target workbook to copy to */
+    private final Workbook targetWorkbook;
+
+    /**
+     * Creates a new PivotStyleCopier instance.
+     *
+     * @param sourceWorkbook the source workbook to copy from
+     * @param targetWorkbook the target workbook to copy to
+     */
+    public PivotStyleCopier(Workbook sourceWorkbook, Workbook targetWorkbook) {
```
RobertBuczek, line 27: Would not be a good idea to do the checks inside the constructor? So that you will have just the logic under the copyPitot:
```
public void copyPivotStyles() {
        var srcStyleSheet = sourceWorkbook.getStylesSource().getCTStylesheet();
        var targetStyleSheet = targetWorkbook.getXSSFWorkbook().getStylesSource().getCTStylesheet();

        if (!srcStyleSheet.isSetDxfs()) {
            return;
        }

        var srcDxfs = srcStyleSheet.getDxfs();
        if (null == srcDxfs || srcDxfs.sizeOfDxfArray() == 0) {
            return;
        }

        var targetDxfs = targetStyleSheet.isSetDxfs() 
            ? targetStyleSheet.getDxfs() 
            : targetStyleSheet.addNewDxfs();

        srcDxfs.getDxfList().forEach(dxf -> {
            var newDxf = targetDxfs.addNewDxf();
            newDxf.set(dxf);
        });

        targetDxfs.setCount(targetDxfs.sizeOfDxfArray());
    }
```

RobertBuczek (dismissed):

src/main/java/uk/gov/laa/gpfd/utils/WorkbookOperations.java
```diff
@@ -49,6 +64,16 @@ default void transferSheet(Workbook sourceWorkbook, Workbook targetWorkbook, Str
...
     }
 
+    /**
+     * Transfers pivot table styles from a source workbook to a target workbook.
+     *
+     * @param sourceWorkbook the workbook containing the original pivot tables
+     * @param targetWorkbook the workbook that has the copied pivot tables
+     */
+    default void copyPivotStyles(Workbook sourceWorkbook, Workbook targetWorkbook) {
+        COPY_PIVOT_STYLES.accept(createPivotFormatCopier().apply(sourceWorkbook, targetWorkbook));
```
HettieS, line 74: be good to look at this line together!

HettieS (approved):
Approved because looks good to me and I saw the working output

src/main/java/uk/gov/laa/gpfd/services/excel/copier/PivotStyleCopier.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        var srcDxfs = srcStyleSheet.getDxfs();
+
+        if (null == srcDxfs || srcDxfs.sizeOfDxfArray() == 0) {
+            // No pivot styles on the source sheet
+            return;
+        }
+
+        var targetDxfs = targetStyleSheet.isSetDxfs() ? targetStyleSheet.getDxfs() : targetStyleSheet.addNewDxfs();
```
RobertBuczek, line 69: what is Dxfs? May we comment what it is and why we are checking NewDxfs?

src/test/java/uk/gov/laa/gpfd/services/excel/copier/PivotStyleCopierTest.java
```diff
@@ -30,68 +48,62 @@ static void setupSourceWorkbookToHavePivotStyles() {
...
+        try (var sourceWorkbook = new XSSFWorkbook();
+             var targetWorkbook = new SXSSFWorkbook()) {
 
-        var pivotFormatCopier = new PivotStyleCopier(sourceWorkbook, targetWorkbook);
-        pivotFormatCopier.copyPivotStyles();
+            var pivotFormatCopier = new PivotStyleCopier(sourceWorkbook, targetWorkbook);
+            pivotFormatCopier.copyPivotStyles();
 
-        assertFalse(targetWorkbook.getXSSFWorkbook().getStylesSource().getCTStylesheet().isSetDxfs());
+            assertFalse(targetWorkbook.getXSSFWorkbook().getStylesSource().getCTStylesheet().isSetDxfs());
```
H3ll3m4, line 74: ditto

src/main/java/uk/gov/laa/gpfd/services/excel/copier/PivotStyleCopier.java
```diff
@@ -0,0 +1,78 @@
...
+        var targetStyleSheet = targetWorkbook.getXSSFWorkbook().getStylesSource().getCTStylesheet();
+
+        if (!srcStyleSheet.isSetDxfs()) {
+            // No pivot styles on the source sheet
+            return;
+        }
+
+        var srcDxfs = srcStyleSheet.getDxfs();
+
+        if (null == srcDxfs || srcDxfs.sizeOfDxfArray() == 0) {
```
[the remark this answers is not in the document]
ChrisCookOC: The return on the apache poi getDxfs method explicitly returns null if the property isn't defined in the style sheet
`return target == null ? null : target;`

it would be empty if the property has been defined but has no values inside it. E.g. maybe the spreadsheet used to have defined styles but the pivot that used them has been deleted so the styles got deleted

src/main/java/uk/gov/laa/gpfd/services/excel/copier/PivotStyleCopier.java
ChrisCookOC: Have added a comment about addNewDxfs and updated the comments further up in the routine documentation and the class documentation that discusses what a dxf is

src/test/java/uk/gov/laa/gpfd/services/excel/copier/PivotStyleCopierTest.java
ChrisCookOC: Have put a comment above the first one, but a more detailed explanation: 


The idea is if it copied something, `isSetDxfs` would return true. 

The XML (and hence the ApachePOI code) has a two-pronged structure. Like some of the other areas of the code, we have to create a container array (Dxfs) to put the individual styles in (Dxf). So a very simplified idea is
```
<dxfs>
   <dxf>(style 1 here)</dxf>
   <dxf>(style 2 here)</dxf>
</dxfs>   
```

When we want to copy styles from the source to the target, the first step is to make that dxfs array, and then add the individual styles. 
`isSetDxfs` is checking the existence of that array. So if it is false we haven't tried to add any styles as the array that contains them is not defined.

So I am testing that if the source has no styles, we have not attempted to create any styles on the target.

H3ll3m4 (approved):
🚀

PR #280, opened 2025-08-04
Feature/lpf 940 fix circular dependency with a ts

SineadCummings:
JIRA: [JIRA ticket number](https://dsdmoj.atlassian.net/browse/LPF-XXX)

## Description of changes
- First provide some context to the changes
- Then explain the changes made, how we do this, and why we choose this implementation over another one.

## Evidence

## Checklist
- [ ] Clean commit history

### DIFF
--- a/.github/checkout-build-openapi/action.yml
+++ b/.github/checkout-build-openapi/action.yml
@@ -17,6 +17,7 @@ runs:
         repository: ministryofjustice/payforlegalaid-openapi
         path: spec
         ssh-key: ${{ env.SOCKET_KEY }}
+        ref: refs/tags/v0.0.3
 
     - name: Build dependency
       run: |
--- a/.github/workflows/feature-build.yml
+++ b/.github/workflows/feature-build.yml
@@ -12,10 +12,10 @@ jobs:
       - name: Checkout GitHub repository
         uses: actions/checkout@v4
 
-      - name: Build and test
-        uses: './.github/build-and-test'
-        env:
-          SOCKET_KEY: ${{ secrets.DEV_API_SOCKET_KEY }}
+#      - name: Build and test
+#        uses: './.github/build-and-test'
+#        env:
+#          SOCKET_KEY: ${{ secrets.DEV_API_SOCKET_KEY }}
 
 #      - name: Snyk Scan
 #        uses: './.github/snyk-scan'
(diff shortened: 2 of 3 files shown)


.github/run_acceptance_tests/action.yml
```diff
@@ -25,7 +29,7 @@ runs:
     - name: Build source code
       run: |
         cd sourceCode
-        mvn -B -DskipTests clean install -s .github/settings.xml -Dmaven.repo.local=${{ github.workspace }}/.m2/repository
+        mvn -B -DskipTests clean install -Dpayforlegalaid-openapi.version=${{ steps.get_openapi_version.outputs.version }}
```
RobertBuczek, line 32: To optimise the build process, I wonder if you would mind considering adding caching for Maven dependencies. This can significantly speed up subsequent builds by reusing downloaded dependencies rather than fetching them each time. GitHub Actions supports this with the actions/cache action.

.github/checkout-build-openapi/action.yml
```diff
@@ -11,16 +16,24 @@ runs:
         java-version: '17'
         cache: 'maven'
 
-    - name: Checkout OpenAPI repository
+    - name: Get openapi version from pom
+      id: get_openapi_version
+      run: |
+        VERSION=$(mvn help:evaluate -Dexpression=payforlegalaid-openapi.version -q -DforceStdout)
```
RobertBuczek, line 22:
```
VERSION=$(mvn help:evaluate -Dexpression=payforlegalaid-openapi.version -q -DforceStdout)
````
this command will always return exit code 0 even in failure scenarios, which leads to a situation where build can continue with empty/malformed version values, You can try with:
```
 mvn help:evaluate -Dexpression=foo-bar -q -DforceStdout
```
this gives you return code 0 which is not right since you we are not using foo-bar.

File: .github/checkout-build-openapi/action.yml [hunk 3]
```diff
@@ -11,16 +16,24 @@ runs:
...
+        echo "version=$VERSION" >> "$GITHUB_OUTPUT"
```
H3ll3m4: @RobertBuczek  sent me a message this morning about the risks with Log4j. What do you think and is there a way to alleviate it?

ChrisCookOC (approved):

PR #291, opened 2025-09-04
LPF-1008 - always pull liquibase image as we are using latest tag

ChrisCookOC:
JIRA: [LPF-1008](https://dsdmoj.atlassian.net/browse/LPF-1008)

## Description of changes
- The correct behaviour for a Kubernetes deployment when using an image with the `latest` tag is to use the `imagePullPolicy: Always`. However, the deployment was created initially with a specified tag rather than `latest` and so the `imagePullPolicy` is defaulted to `IfNotPresent`. 
- As such, it is prudent to manually set the `imagePullPolicy` so that none of our deployments rely on an incorrect default.
- On a practical note, `imagePullPolicy: Always` means that each time the pod is created, it will go to ECR, get the digest (this is the sha of the image) of the image tagged `latest`, and only download it if said image digest is not already in it's cache. In this sense we should still only be transferring the image from ECR to Kubernetes once, and I wouldn't expect a notable increase in the amount of image transferring happening (it is not like every time we redeploy or restart a pod, it will download the image!)
- Snyk - netty issue. It's a transitive dependency and the upgrade has not been enacted by either AWS or Azure yet. Note that since we use tomcat in Spring and we use the sync (rather than async) S3 client, we do not use netty currently. 

## Checklist
- [ ] Clean commit history

[LPF-1008]: https://dsdmoj.atlassian.net/browse/LPF-1008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -2,10 +2,12 @@
 version: v1.25.1
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
 ignore:
-  SNYK-JAVA-IONETTY-11799531:
+  SNYK-JAVA-IONETTY-12485149:
     - '*':
         reason: >-
-          Azure Spring Integration has yet to upgrade their versions with the new netty-handler.
-        expires: 2025-08-27T09:00:00.000Z
-        created: 2025-08-14T16:51:23.725Z
+          Neither Azure nor S3 have upgraded yet. We don't actively use netty as
+          a server or use async s3 calls so we do not run any netty processes at
+          present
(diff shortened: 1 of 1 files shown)


HettieS (approved):
Looks good, I guess let's just keep an eye if we think anything is downgraded from pulling the image too often. Not quite sure how we'll track that though!

PR #438, opened 2026-04-14
LPF-1127 - report tracking

ChrisCookOC:
JIRA: [LPF-1127](https://dsdmoj.atlassian.net/browse/LPF-1127)

## Description of changes
- New report Tracking Table in an RDS database
- RDS here is both a PoC and shows it is possible to decouple us from MoJFIN which will be deprecated
- RDS allows us to store User ID which falls under PII
- No change to existing tables using MoJFIN
- New database uses Flyway
- When user downloads a report (via `/csv/{id}`, `/excel/{id}` or `/reports/{id}/file`) it will log an entry in the tracking table with their User ID, the report ID and the timestamp
- Remove unused code around fetching Tokens from Auth - doesn't apply in SILAS world
- Stop the `ReportsControllerTest` and `PolicyControllerTest` loading up the entire app. Unnecessary for unit tests and causes setup overhead.
- Move the `OAuth2LoginTests` - they are duplicates of tests already covered by the ITs and they were loading the entire app so acting like ITs anyway
- Docker compose creates a postgres db locally

###V2
- Ensure all three responses use a `StreamingResponseBody` for the data stream and wrap that in a try/catch so we can monitor its success or lack thereof. 
- We now only enter a report tracking row if that stream is successfully. Note that a stream being successful means all the server stuff is done and the client can download everything, not that the client has downloaded everything or the download will fully succeed
- Previously the HTTP Response was built down in the depths of the `StreamingService`, which wasn't a great place to start manipulating the stream or tracking the report status. So I've separated out the concerns. The `StreamingService` now just returns us the data stream. The HTTP Response is built by a `ResponseBuilder`.
- Tracking Insert is now Async so the result of that does not affect the user

## Evidence
Video link shared in Slack
image

## Checklist
- [ ] Clean commit history

[LPF-1127]: https://dsdmoj.atlassian.net/browse/LPF-1127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pom.xml
```diff
@@ -432,10 +444,6 @@
             <artifactId>spring-boot-starter-actuator-test</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
```
ChrisCookOC, line 435: This is a duplicate we already have it elsewhere in this file

src/it/java/uk/gov/laa/gpfd/integration/BaseIT.java
```diff
@@ -6,55 +6,86 @@
...
+    @Autowired
+    MockMvc mockMvc;
+
+    static final PostgreSQLContainer POSTGRES = new PostgreSQLContainer("postgres:18")
+            .withDatabaseName("glad")
+            .withUsername("postgres")
+            .withPassword("password")
+            .withExposedPorts(5432);
+
+    static {
```
DevarshBarotOC: Seeing as we're running these tests using `@TestContainers` and we're using Spring 4, this can be refactored to remove all this boilerplate wiring:

```
@Testcontainers
@SpringBootTest
class MyIntegrationTest {

    @Container
    @ServiceConnection
    static PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>("postgres:18");

    @Autowired
    private MockMvc mockMvc;
}
```

`@ServiceConnection` basically does the database config for you and plugs it into spring without having to manually configure it like above. More importantly, it gets rid of this static block which would otherwise cause issues with the testcontainers lifecycle management. It may also display weird parallel test behaviour and could start the container eagerly even if the tests don't run.

src/it/java/uk/gov/laa/gpfd/integration/BaseIT.java
```diff
@@ -6,55 +6,86 @@
...
+        r.add("gpfd.datasource.tracking.jdbcUrl", POSTGRES::getJdbcUrl);
+        r.add("gpfd.datasource.tracking.username", POSTGRES::getUsername);
+        r.add("gpfd.datasource.tracking.password", POSTGRES::getPassword);
+
+        r.add("spring.flyway.enabled", () -> true);
+        r.add("spring.flyway.url", POSTGRES::getJdbcUrl);
+        r.add("spring.flyway.user", POSTGRES::getUsername);
+        r.add("spring.flyway.password", POSTGRES::getPassword);
+
+    }
```
DevarshBarotOC, line 54: If you use the suggestion above to use `@ServiceConnection`, this can also be removed.

src/main/java/uk/gov/laa/gpfd/utils/SecurityUtils.java
```diff
@@ -39,12 +43,27 @@ public List<String> extractRoles() {
...
+        }
+
+        return switch (auth.getPrincipal()) {
+            case OidcUser user -> user.getAttribute("oid");
+            case null -> throw new PrincipalIsNullException();
+            default -> throw new UnexpectedAuthClassException("Unexpected Auth Type: " + auth.getClass().getName());
+        };
+
+    }
+
```
DevarshBarotOC, line 72: The way this is written isn't bad but it is a little brittle. Refactoring it as such removes the need to do `UUID.fromString(securityUtils.extractUserId());` in the `ReportTrackingDao`. This also removes a hidden dependency on security context in service classes which isn't visible in the signatures. 

```
public UUID extractUserId() {
    Authentication auth = SecurityContextHolder.getContext().getAuthentication();

    if (auth == null || auth.getPrincipal() == null) {
        throw new AuthenticationIsNullException();
    }

    return switch (auth.getPrincipal()) {
        case OidcUser user -> {
            String oid = user.getAttribute("oid");
            if (oid == null) throw new IllegalStateException("Missing oid");
            yield UUID.fromString(oid);
        }
        default -> throw new UnexpectedAuthClassException(
            "Unsupported principal: " + auth.getPrincipal().getClass().getName()
        );
    };
}
```

In an ideal world, you could remove the security context from `ReportTrackingDao` and move the logic to the controller - so effectively you would be doing, for example:

```
@Override
public ResponseEntity<StreamingResponseBody> getExcelById(
        @AuthenticationPrincipal OidcUser user,
        UUID id) {

    UUID userId = UUID.fromString(user.getAttribute("oid"));

    log.info("Returning an Excel report for id {} to user {}", id, userId);

    // Enforce role-based access control for this report
    reportDao.verifyUserCanAccessReport(id);

    // Validate format before attempting to stream
    reportManagementService.validateReportFormat(id, XLSX);

    // Track access
    reportTrackingDao.insertTrackingRow(userId, id);

    return streamingService.stream(id, XLSX);
}
```
And then in the DAO, you could remove SecurityUtils from the constructor, and the UserId could be passed into the DAO as a UUID instead. Therefore:
- The controller handles the authentication side, as it should
- We safely extract the userId in the format we want, thus removing the need for additional parsing from string etc.
- We also remove the hidden security context dependency from the ReportTrackingDao which makes it more deterministic and testable.

src/main/java/uk/gov/laa/gpfd/utils/TokenUtils.java
```diff
@@ -1,56 +1,17 @@
...
-import org.springframework.security.oauth2.core.user.OAuth2User;
-import org.springframework.security.oauth2.jwt.Jwt;
-import uk.gov.laa.gpfd.exception.UnableToGetAuthGroupException.AuthenticationIsNullException;
-import uk.gov.laa.gpfd.exception.UnableToGetAuthGroupException.PrincipalIsNullException;
-import uk.gov.laa.gpfd.exception.UnableToGetAuthGroupException.UnexpectedAuthClassException;
 
-import java.util.List;
 import java.util.UUID;
 
 public abstract class TokenUtils {
```
DevarshBarotOC: Since this is no longer a TokenUtils class as we have deleted all the token util logic, it could be worth refactoring this to:

```
public final class ReportIds {

    private ReportIds() {}

    public static final UUID REP000 = UUID.fromString("523f38f0-2179-4824-b885-3a38c5e149e8");
    public static final UUID REP012 = UUID.fromString("cc55e276-97b0-4dd8-a919-26d4aa373266");
    public static final UUID REP013 = UUID.fromString("aca2120c-8f82-45a8-a682-8dedfb7997a7");
    public static final UUID REP014 = UUID.fromString("55daf3c1-28f0-4260-9396-2ee6d537abab");
}
```

src/main/java/uk/gov/laa/gpfd/dao/ReportTrackingDao.java
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.laa.gpfd.exception.DatabaseWriteException;
+import uk.gov.laa.gpfd.utils.SecurityUtils;
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.UUID;
+
+@Slf4j
+@Service
+public record ReportTrackingDao(@Qualifier("trackingJdbcTemplate") JdbcOperations trackingJdbcTemplate, SecurityUtils securityUtils) {
```
DevarshBarotOC: To be refactored to remove this security context dependency.

src/main/java/uk/gov/laa/gpfd/dao/ReportTrackingDao.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private static final String insertSql = "INSERT INTO GLAD.REPORT_TRACKING(ID, REPORT_ID, USER_ID, DOWNLOAD_TIME) VALUES (?, ?, ?, ?)";
+
+    public void insertTrackingRow(UUID reportId) {
```
DevarshBarotOC: To be refactored to pass a userID from the controller instead.

src/test/java/uk/gov/laa/gpfd/dao/ReportTrackingDaoTest.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class ReportTrackingDaoTest {
+
+    @Mock
+    JdbcOperations trackingJdbcTemplate;
+
+    @Mock
```
DevarshBarotOC: Tests will need refactoring once changes have been made to ReportTrackingDao

DevarshBarotOC (changes requested):

src/it/java/uk/gov/laa/gpfd/integration/BaseIT.java
masuk-kazi98: what he said ^

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -77,8 +77,9 @@ public ResponseEntity<StreamingResponseBody> csvIdGet(UUID requestedId) {
 
         // Validate that this report is actually a CSV report
         reportManagementService.validateReportFormat(requestedId, CSV);
-
-        return streamingService.stream(requestedId, CSV);
+        var response = streamingService.stream(requestedId, CSV);
+        reportTrackingDao.insertTrackingRow(requestedId);
+        return response;
```
H3ll3m4, line 84: If the tracking fail, download fails too while the report could actually be ready? Would that mean that report download depends on the tracking DB insert? Would that be better to make sure we stream and keep the tracking best effort?

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -115,7 +116,9 @@ public ResponseEntity<StreamingResponseBody> getExcelById(UUID id) {
         // Validate format before attempting to stream
         reportManagementService.validateReportFormat(id, XLSX);
 
-        return streamingService.stream(id, XLSX);
+        var response = streamingService.stream(id, XLSX);
+        reportTrackingDao.insertTrackingRow(id);
+        return response;
```
H3ll3m4, line 123: same

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -125,7 +128,9 @@ public ResponseEntity<InputStreamResource> getReportDownloadById(UUID id) {
         // Validate that this report is S3STORAGE format
         reportManagementService.validateReportFormat(id, S3STORAGE);
 
-        return fileDownloadService.getFileStreamResponse(id);
+        var response = fileDownloadService.getFileStreamResponse(id);
+        reportTrackingDao.insertTrackingRow(id);
+        return response;
```
H3ll3m4, line 135: same?

src/main/java/uk/gov/laa/gpfd/dao/ReportTrackingDao.java
H3ll3m4: Does it need to be a record when it is a service? This class do things rather holding values?

src/main/java/uk/gov/laa/gpfd/dao/ReportTrackingDao.java
H3ll3m4: Do we want to validate reportId?

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -77,8 +77,9 @@ public ResponseEntity<StreamingResponseBody> csvIdGet(UUID requestedId) {
 
         // Validate that this report is actually a CSV report
         reportManagementService.validateReportFormat(requestedId, CSV);
-
-        return streamingService.stream(requestedId, CSV);
+        var response = streamingService.stream(requestedId, CSV);
```
H3ll3m4, line 82: We prepare the stream here but we don't download yet as it is asynchronous. What happen if the stream download fails as we have tracked the download. Do we rollback?

src/main/resources/flyway/migration/schema/V1__create_report_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE report_tracking (
```
H3ll3m4, line 1: Are we using flyway rather than liquibase?

src/test/java/uk/gov/laa/gpfd/dao/ReportTrackingDaoTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    void insertTrackingRow_shouldUpdateDbWithCorrectValues() {
+        var reportId = UUID.randomUUID();
+        var userId = UUID.randomUUID();
+        when(securityUtils.extractUserId()).thenReturn(userId.toString());
+        reportTrackingDao.insertTrackingRow(reportId);
+
+        verify(trackingJdbcTemplate).update(anyString(),any(UUID.class), eq(reportId), eq(userId), any(Timestamp.class));
```
H3ll3m4: anyString() should be the right command`insert into...`

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
ChrisCookOC: Happy to do it that way if we want. My thinking was as the purpose of this leans more towards like auditing access to PII than just collecting stats on downloads (we have the logs/ingress for that), it makes sense to require it to happen. But yeah can also see argument against it. So can take it whatever way the team/product wants :)

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
ChrisCookOC: So, this is an interesting question as to what it means and what can we track. Because once we send a stream to the browser - certainly for the big S3 files like the 700MB one - we have no guarantee it downloaded. For example I had an issue downloading the test file over 1GB and my browser would kill the download but our service never knew, or I could stop and restart the download over and over and the service was seemingly none the wiser?

It looks like we could perhaps change our response types, and like create a custom extension of the response stream that we can track the bytes sent? Still doesn't guarantee the user received the bytes or the browser saved them though. 

Happy to hear any suggestions / thoughts / discuss how we can actually tell something is downloaded though and what like our threshold for "it being downloaded is". Because from my research I think there's always some element of like uncertainty there. :/

src/main/resources/flyway/migration/schema/V1__create_report_table.sql
ChrisCookOC: Yep we discussed trying out Flyway for the Postgres database [here](https://mojdt.slack.com/archives/C07UK3ZDPPE/p1773758742796029?thread_ts=1773757393.514559&cid=C07UK3ZDPPE), because its simpler (the complexity of Liquibase is perfect for use with Oracle DB specifically but less so elsewhere) and brings us inline with the DCRS/Data Stewardship use of Postgres.

src/it/java/uk/gov/laa/gpfd/integration/BaseIT.java
ChrisCookOC: had some discussion with dev about this. have made some clean up here but we can't fully utilise the ServiceConnection due to the way we use our multiple datasources right now.

src/main/java/uk/gov/laa/gpfd/utils/SecurityUtils.java
ChrisCookOC: Updated, cheers Dev

src/main/java/uk/gov/laa/gpfd/dao/ReportTrackingDao.java
ChrisCookOC: Moved to a class. all our other DAOs currently are records, but I think its just to make Dependency Injection of the variable easier since that `@Qualifier` can go nicely in the constructor of a record.

src/main/java/uk/gov/laa/gpfd/dao/ReportTrackingDao.java
ChrisCookOC: I'm not convinced we do here, just from a flow. The controller has already validated the ID and the generation process will also validate the ID, so adding another database call to validate the report ID here feels like unnecessary repetition to me? Happy to discuss further

src/it/java/uk/gov/laa/gpfd/integration/ReportGetFileIT.java
```diff
@@ -75,54 +63,29 @@ final class ReportGetFileIT extends BaseIT {
...
-                    .andExpect(status().isOk())
-                    .andExpect(content().contentType(APPLICATION_OCTET_STREAM))
-                    .andExpect(header().longValue("Content-Length", 25L))
-                    .andExpect(header().string("Content-Disposition", "attachment; filename=\"report_2025-12-15.csv\""))
-                    .andExpect(content().string("csv,data,here,123,4.3,cat"));
-        }
-
     @Test
     @SneakyThrows
-    void shouldCloseStreamWhenSuccessful() {
```
ChrisCookOC, line 106: Test is no longer valid.

When we were using `InputStreamResource` as our response type, we controlled the full stream including invoking closure.

Now, we delegate to a `StreamingResponseBody` for which SpringBoot is responsible for lifecycle including closing

src/test/java/uk/gov/laa/gpfd/controller/ReportsControllerTest.java
```diff
@@ -209,40 +260,10 @@ void downloadCsvRejectsInvalidFiletypes(String reportId, String actualFormat) th
                 .stream(uuid, FileExtension.CSV);
     }
 
-    @Test
-    void downloadCsvSucceedsForCsvReport() throws Exception {
```
ChrisCookOC, line 212: Duplicate scenario of `downloadCsvReturnsCorrectResponse`

src/test/java/uk/gov/laa/gpfd/controller/ReportsControllerTest.java
```diff
@@ -299,40 +326,49 @@ void getReportDownloadByIdRejectsInvalidFiletypes(String reportId, String actual
     }
 
     @Test
-    void getReportDownloadByIdSucceedsForS3StorageReport() throws Exception {
```
ChrisCookOC, line 301: Duplicate scenario of `getReportDownloadByIdReturnsCorrectResponseEntity`

H3ll3m4 (approved):
🚀

masuk-kazi98 (approved):

HettieS (approved):

PR #450, opened 2026-04-28
LPF-1357: Added Gatling performance tests

masuk-kazi98:
JIRA: [LPF-1357](https://dsdmoj.atlassian.net/browse/LPF-1357)

## Description of changes
- Created Gatling simulation tests for performance of API
- new manual workflow added to run perf tests

## Evidence
[concurrencysimulation-20260428193027599.zip](https://github.com/user-attachments/files/27178981/concurrencysimulation-20260428193027599.zip)
[reportdownloadsimulation-20260428191643322.zip](https://github.com/user-attachments/files/27178982/reportdownloadsimulation-20260428191643322.zip)
[reportlistsimulation-20260428192247643.zip](https://github.com/user-attachments/files/27178983/reportlistsimulation-20260428192247643.zip)

## Checklist
- [ ] Clean commit history

[LPF-1357]: https://dsdmoj.atlassian.net/browse/LPF-1357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ Acceptance tests are located [here](https://github.com/ministryofjustice/payforl
 
 ### Performance Tests
 
-Performance tests have been set up in payforlegalaid using Gatling to things like retrieving the list of reports, downloading
+Performance tests have been set up in payforlegalaid using Gatling to evaluate things like retrieving the list of reports, downloading
 the reports and also concurrency.
 
 src/test/java/uk.gov.laa.gpfd/simulations
--- a/README.md
+++ b/README.md
@@ -180,6 +180,25 @@ test/resources folder.
 
 Acceptance tests are located [here](https://github.com/ministryofjustice/payforlegalaid-tests).
 
+### Performance Tests
+
+Performance tests have been set up in payforlegalaid using Gatling to things like retrieving the list of reports, downloading
+the reports and also concurrency.
+
+src/test/java/uk.gov.laa.gpfd/simulations
+
+The base URL that the tests use is set within gatling.properties within /resources. Set to UAT env by default.
+To run the tests, first log in manually into desired environment, then export session cookie `JSESSIONID`:
+
+`export JSESSIONID=<cookie>`
+
+and then run:
+
+`mvn gatling:test -Dgatling.simulationClass=uk.gov.laa.gpfd.simulations.<test-name> -Dmaven.antrun.skip=true -Dmaven.resources.skip=true`
+
+Replace <test-name> with the choice of test. Reports are generated automatically by Gatling and can be accessed at the end of a
+test run within the terminal
+
 ## CI/CD
 
 GitHub Actions is used for CI/CD.
--- a/README.md
+++ b/README.md
@@ -182,22 +182,29 @@ Acceptance tests are located [here](https://github.com/ministryofjustice/payforl
 
 ### Performance Tests
 
-Performance tests have been set up in payforlegalaid using Gatling to evaluate things like retrieving the list of reports, downloading
-the reports and also concurrency.
+Performance tests have been set up in payforlegalaid using Gatling to evaluate things like retrieving the list of reports, 
+downloading the reports and also concurrency.
 
 src/test/java/uk.gov.laa.gpfd/simulations
 
 The base URL that the tests use is set within gatling.properties within /resources. Set to UAT env by default.
-To run the tests, first log in manually into desired environment, then export session cookie `JSESSIONID`:
+To run the tests, first log in manually into desired environment, then copy the session cookie, `JSESSIONID`, from
+within the browser DevTools -> Application -> Cookies, then run:
 
-`export JSESSIONID=<cookie>`
-
-and then run:
-
-`mvn gatling:test -Dgatling.simulationClass=uk.gov.laa.gpfd.simulations.<test-name> -Dmaven.antrun.skip=true -Dmaven.resources.skip=true`
+```bash
+export JSESSIONID=<insert session id>
+mvn gatling:test -Dgatling.simulationClass=uk.gov.laa.gpfd.simulations.<test-name> -Dmaven.antrun.skip=true -Dmaven.resources.skip=true`
+```
 
 Replace <test-name> with the choice of test. Reports are generated automatically by Gatling and can be accessed at the end of a
-test run within the terminal
+test run within the terminal.
+
+Alternatively, developers can trigger the manual GitHub Actions workflow at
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -180,6 +180,25 @@ test/resources folder.
...
 
+### Performance Tests
+
+Performance tests have been set up in payforlegalaid using Gatling to things like retrieving the list of reports, downloading
+the reports and also concurrency.
+
+src/test/java/uk.gov.laa.gpfd/simulations
+
+The base URL that the tests use is set within gatling.properties within /resources. Set to UAT env by default.
+To run the tests, first log in manually into desired environment, then export session cookie `JSESSIONID`:
```
H3ll3m4: Can you detail how to get the session cookie?

README.md
```diff
@@ -180,6 +180,25 @@ test/resources folder.
 
 Acceptance tests are located [here](https://github.com/ministryofjustice/payforlegalaid-tests).
 
+### Performance Tests
+
+Performance tests have been set up in payforlegalaid using Gatling to things like retrieving the list of reports, downloading
```
H3ll3m4: Missing a word here like test or evaluate?

src/test/java/uk/gov/laa/gpfd/simulations/ConcurrencySimulation.java
```diff
@@ -0,0 +1,57 @@
...
+            .exec(
+                    http("View report detail [#{size}]")
+                            .get("/reports/#{id}")
+                            .check(status().in(200, 429))
+            )
+            .pause(1, 3)
+            .exec(
+                    http("Download [#{format}] [#{size}]")
+                            .get("/reports/#{id}/#{format}")
+                            .check(status().in(200, 429, 500, 503))
```
H3ll3m4, line 52: Can you detail how did you choose those error codes? Are we testing 500 and 503 that are failures?

H3ll3m4:
I wonder if that would be useful to be able to trigger those tests with a workflow?

H3ll3m4:
Should we be asserting the low performances programmatically? Like response time < 500 ms and error rates < 0.1%?

H3ll3m4:
I'm not clear about the test performed, was it UI or API included?

README.md
masuk-kazi98: it's detailed enough no?😅

masuk-kazi98:
> I'm not clear about the test performed, was it UI or API included?

Just the API in this PR, UI is tested in the pfla-tests PR

masuk-kazi98:
> Should we be asserting the low performances programmatically? Like response time < 500 ms and error rates < 0.1%?

We can do yes, PR was just an initial pass and observation on performance just to see what kind of results we were getting. Right now we already have:

P95 response time assertion (10s in one scenario)
global failure rate threshold

We can probs improve this by:

tightening SLAs per endpoint (e.g. < 500ms for list endpoints if that’s realistic)
splitting assertions by scenario rather than global only
defining thresholds based on expected production behaviour

I’ve kept them relatively loose initially as we’re still establishing baseline performance under load, but we can absolutely evolve this into stricter SLO-style assertions.

Do we have SLO's? maybe from other teams? that we can test against

masuk-kazi98:
> I wonder if that would be useful to be able to trigger those tests with a workflow?

Personally i don't think it adds much value having this run after each PR/merge. This should be something we run from time to time to assess and monitor our service, maybe monthly/weekly to ensure performance remains the same through additions of new features/more reports/larger reports/more users etc.

src/test/java/uk/gov/laa/gpfd/simulations/ConcurrencySimulation.java
masuk-kazi98: the intent here isn’t to “test for failures”, but to observe system behaviour under load rather than assume a perfectly clean response model.

200 = expected successful path
429 = intentionally included to capture rate limiting under load
500 / 503 = included to surface true instability under stress

so in the ConcurrencySimulation specifically, the goal is:

“What happens when we push the system — does it degrade gracefully (429), or break (5xx)?”

Whereas in the more targeted scenarios (download/list simulations), I’ve tightened expectations to only 200 to represent normal user journeys

README.md
ChrisCookOC: Are we meant to be running these against cloud platform environments. I assume they are low enough intensity/frequency we can get away with it?

pom.xml
```diff
@@ -34,6 +34,9 @@
         <maven-dependency-plugin.version>3.10.0</maven-dependency-plugin.version>
         <maven-antrun-plugin.version>3.2.0</maven-antrun-plugin.version>
         <maven-war-plugin.version>3.5.1</maven-war-plugin.version>
+
+        <gatling.version>3.15.0</gatling.version>
+        <gatling-maven-plugin.version>4.15.0</gatling-maven-plugin.version> <!-- Latest version -->
```
ChrisCookOC: can we remove the comment saying latest version?

README.md
masuk-kazi98: shouldn't be an issue, have ramped them down enough to not be hitting 429s constantly so shouldn't come up as weird behaviour on CP's end. It's good to run perf tests as close to the prod env as possible to replicate real user behaviour and service response from our pods. we don't really have a Staging env or else would be using that, UAT next best thing

ChrisCookOC (dismissed):

README.md
H3ll3m4: I would say the browser session cookie you can find on the developer tool console on the browser bla bla ba export using the command bla bla bla

src/test/java/uk/gov/laa/gpfd/simulations/ConcurrencySimulation.java
H3ll3m4: Cool do we want to add just a comment in the code what is the simulation doing?

H3ll3m4:
> > I wonder if that would be useful to be able to trigger those tests with a workflow?
> 
> Personally i don't think it adds much value having this run after each PR/merge. This should be something we run from time to time to assess and monitor our service, maybe monthly/weekly to ensure performance remains the same through additions of new features/more reports/larger reports/more users etc.

Yes if we had a workflow we could trigger it when we want as you say, monthly...

H3ll3m4 (dismissed):
🚀

H3ll3m4 (approved):

HettieS (approved):

ChrisCookOC (approved):

PR #467, opened 2026-05-07
LPF-1189 - handle no roles supplied from SiLAS better

ChrisCookOC:
JIRA: [LPF-1189](https://dsdmoj.atlassian.net/browse/LPF-1189)

## Description of changes
- Currently in the SILAS feature branch, when we get sent no roles from SiLAS (be that due to user or app setup issues), we try and make a database call to fetch all reports available to a user with no roles
- This throws a database exception as it tries to bind empty list to the database request
- There is no valid scenario where a user has no roles. When setting up a user on SiLAS to use GLAD, you **must** give them at least one role. As such we should error straight away if we parse a user's auth and find no roles inside.
## Checklist
- [x] New tests are included if this a new feature
- [x] Tests and linter checks are passing locally
- [x] Documentation README.md & Confluence have been updated
- [x] TODOs, commented code and print traces have been removed
- [x] Any dependant changes have been merged in downstream modules
- [x] Clean commit history

[LPF-1189]: https://dsdmoj.atlassian.net/browse/LPF-1189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/uk/gov/laa/gpfd/exception/UnableToParseAuthDetailsException.java
+++ b/src/main/java/uk/gov/laa/gpfd/exception/UnableToParseAuthDetailsException.java
@@ -70,4 +70,44 @@ public abstract sealed class UnableToParseAuthDetailsException extends RuntimeEx
         }
     }
 
+    /**
+     * Exception throw when we can't get attributes out of the auth token sent via SiLAS.
+     */
+    public static final class NoAttributesOnTokenException extends UnableToParseAuthDetailsException {
+
+        /**
+         * Constructs a new {@code NoAttributesOnTokenException}.
+         */
+        public NoAttributesOnTokenException() {
+            super("Could not parse attributes from token");
+        }
+    }
+
+    /**
+     * Exception throw when user has no roles returned in the token.
+     * This isn't really a valid setup in SiLAS so it suggests SiLAS isn't sending us the roles.
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/laa/gpfd/exception/UnableToParseAuthDetailsException.java
```diff
@@ -70,4 +70,44 @@ public NoOidSetOnTokenException() {
...
+        /**
+         * Constructs a new {@code NoRolesException}.
+         */
+        public NoRolesInAttributeException() {
+            super("No roles were sent in the token");
+        }
+    }
+
+    /**
+     * Exception throw when roles returned in the token but there are none.
```
DevarshBarotOC: I would reword this slightly:

* Exception throw when roles returned in the token are not assigned to the user.

ChrisCookOC: Can I reword your reword 

     * Exception throw when roles key is returned in the token attributes but the user has none assigned.

DevarshBarotOC (approved):

masuk-kazi98 (approved):

PR #474, opened 2026-05-11
feat(LPF-1365): Implement API pattern target configuration for CSRF Protection

masuk-kazi98:
JIRA: [LPF-1365](https://dsdmoj.atlassian.net/browse/LPF-1365)

## Description of changes
- Updated SecurityConfig + SecurityConfigLocal to strengthen CSRF protection
- set session cookie as `same-site: lax` as `strict` is too rigid and would block Entra auth from redirecting correctly
- CSRF set to strict
- Updated application template file to include csrf token
- CsrfConfig class created 
- IT's updated
- Split out security config due to lots of shared/dupe code

## Evidence
Screenshot 2026-05-11 at 17 40 55

## Checklist
- [ ] Title follows the naming {Type}: {TICKET-NUMBER}-{brief-description}. All fields should be in the branch name. Type is the type of change: feature, documentation, bugfix...
- [ ] New tests are included if this a new feature
- [ ] Tests and linter checks are passing locally
- [ ] Documentation README.md & Confluence have been updated
- [ ] TODOs, commented code and print traces have been removed
- [ ] Any dependant changes have been merged in downstream modules
- [ ] Clean commit history

[LPF-1365]: https://dsdmoj.atlassian.net/browse/LPF-1365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #488, opened 2026-05-18
fix(LPF-1189): streamline test setup

ChrisCookOC:
JIRA: [LPF-1189](https://dsdmoj.atlassian.net/browse/LPF-1189)

## Description of changes
Previously we had two test config setups - `test` provided an environment without any auth, and `testauth` added auth to the settings.
Previously, we only used auth on the "higher environments" and not local, so to a degree these two setups mirrored our real setups
In the SiLAS world, we will always have auth even on local. As such the `test` setup is deprecated. And its confusing having two similar but subtly different setups - in some cases we'd even use the `testauth` profile with say the `test` property file, or combined them with the `TestSecurityConfig`, to run some sort of hybrid setup! So I've killed off the `test` setup in favour of just `testauth`.

so the changes to make this happen
- let `BaseIT` setup the config and make sure all ITs use the same basic config (some may have additional configs for specific scenarios they are testing)
- Move some of the tests still relevant from `SecurityConfigLocalIT` to `SecurityConfigIT`, and remove the others that related to an Entra-free world
- Merge `SecurityConfig` and `SecurityConfigLocal`. In the SiLAS branch after the removal of h2 in local, they were identical bar one line!
- Remove `S3ConfigTest`. The tests were just checking beans were created correctly, but in the test setup the beans are different. So to test the "real" beans we would either need very complicated setup in this one test, or to complicate the setup of the other tests that needed the "test" beans. In the end I decided that checking beans exist isn't really a useful test here anyway and killed them off.
- Removed `TestSecurityConfig` - all tests using same config now.

## Evidence
Tests all pass still!
image

## Checklist
- [ ] Clean commit history

[LPF-1189]: https://dsdmoj.atlassian.net/browse/LPF-1189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/uk/gov/laa/gpfd/utils/SecurityUtils.java
```diff
@@ -53,7 +53,9 @@ public List<String> extractRoles() {
                     throw new NoRolesInAttributeException();
                 }
 
+                log.debug("Is the role claim in there? {}", attributes.containsKey(ROLE_CLAIM));
                 Object rawRoles = attributes.get(ROLE_CLAIM);
+                log.debug("Raw roles are {}", rawRoles);
```
DevarshBarotOC: Logging the raw object from the claim could potentially expose sensitive (or unexpected) data depending on the structure of the OIDC attributes..

I would be a lot more comfortable with putting this debug line after: `List<String> parsedRoles = parseRoles(rawRoles);` and doing something like:

```
List<String> parsedRoles = parseRoles(rawRoles);
log.debug("Roles are {}", parsedRoles);
```

This way we aren't exposing the entire object, only the validated and expected roles within the claim.

src/main/java/uk/gov/laa/gpfd/config/SecurityConfig.java
```diff
@@ -88,7 +88,10 @@ SecurityFilterChain staticChain(HttpSecurity http) {
      * @return a configured {@link SecurityFilterChain} object.
      */
     @Bean
-    SecurityFilterChain filterChain(HttpSecurity httpSecurity) {
+    SecurityFilterChain filterChain(HttpSecurity httpSecurity, Environment env) {
+
+        boolean isLocal = Arrays.stream(env.getActiveProfiles()).anyMatch(p -> p.equals("local") || p.equals("testauth"));
```
DevarshBarotOC: Since we're using spring, we could do this slightly differently like so:
`boolean isLocal = env.acceptsProfiles(Profiles.of("local", "testauth"));`

That way we avoid stream logic, reads better and is easier to scale if we add more profiles (for whatever reason) in the future.

DevarshBarotOC (changes requested):

src/main/java/uk/gov/laa/gpfd/utils/SecurityUtils.java
ChrisCookOC: Iv'e just removed it completely as it was only there for me debugging a system where the roles werne't coming through

src/main/java/uk/gov/laa/gpfd/config/SecurityConfig.java
ChrisCookOC: Done, good shout

masuk-kazi98:
Does this PR mean i'll have to rewrite my securityconfig changes on my [PR](https://github.com/ministryofjustice/payforlegalaid/pull/474)

masuk-kazi98 (approved):

H3ll3m4 (approved):
🚀

PR #505, opened 2026-06-01
ci(LPF-000): skip snapshots, update deployment process

masuk-kazi98:
JIRA: [JIRA ticket number](https://dsdmoj.atlassian.net/browse/LPF-XXX)

## Description of changes
- Updated release-please to stop creating snapshot PR's
- Updated deploy.yml
1. release-please PR merges trigger automatic deployments from dev -> prod (no change to existing 'review required')
2. feature branch merges now only trigger automatic deployments to dev ONLY, not beyond
3. created workflow dispatch, so that for testing purposes we can manually deploy changes to: dev, uat, dev+uat, or for hotfixes etc can deploy from dev -> prod. Removes the need to have to keep cancelling workflows before hitting uat/prod to test dev/uat

workflow_diagram

## Evidence

## Checklist
- [ ] Title follows the format `{type}({TICKET-NUMBER}): {brief description}`.   Example: `feat(ABC-123): add caching layer`
  - Type must be one of: 
    - feat
    - fix
    - docs
    - chore
    - refactor
    - test
    - ci
    - build
    - perf
- [ ] Clean commit history

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,9 +42,9 @@ jobs:
           is_release=false
 
           if [[ "$EVENT_NAME" == "push" ]]; then
-            msg="$COMMIT_MSG"
+            msg="${COMMIT_MSG:-}"
 
-            if [[ "$msg" =~ ^chore(\([^)]+\))?:\ release\ [0-9] ]]; then
+            if [[ -n "$msg" ]] && [[ "$msg" == chore\(main\):\ release* ]]; then
               is_release=true
             fi
           fi
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -35,6 +35,7 @@ jobs:
 
     steps:
       - id: check
+        name: Check if deployment is a release merge
         env:
           EVENT_NAME: ${{ github.event_name }}
           COMMIT_MSG: ${{ github.event.head_commit.message }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,7 +22,6 @@ on:
           - uat
           - all
 
-# One deployment at a time on main; queue later runs instead of cancelling mid-deploy.
 concurrency:
   group: deploy-${{ github.ref }}
   cancel-in-progress: false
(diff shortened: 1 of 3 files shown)


DevarshBarotOC (dismissed):

.github/workflows/deploy.yml
```diff
@@ -5,10 +5,77 @@ on:
...
 
 jobs:
+  release_check:
+    name: Determine deployment path
+    runs-on: ubuntu-latest
+    outputs:
+      is_release_merge: ${{ steps.check.outputs.is_release_merge }}
+
+    steps:
+      - id: check
```
ChrisCookOC, line 37: Can you give the step a name. Currently in the runner it appears as 

Image

which is confusing

.github/workflows/deploy.yml
```diff
@@ -5,10 +5,77 @@ on:
...
+        description: >-
+          For feature branches / ad-hoc deploys (merges to main do not auto-deploy except release-please).
+          dev — development only.
+          dev-uat — development then UAT (UAT requires approval).
+          uat — UAT only (does not deploy dev first).
+          all — development, UAT, then prod.
+        type: choice
+        required: true
+        default: dev
+        options:
```
ChrisCookOC, line 19: I've spent some time trying to bypass this. I think if we set this workflow to be reusable you could input an invalid option, but its not setup that way so its fine. Github Cli seems to enforce it is one of the options listed so that is good.

Anyway, doesn't matter really, what I was really wondering/trying to prove/disprove is do we need to validate that the input does match these options before we use it later on, in case I've sent in an option of `dev; rm -rf` or something?

masuk-kazi98: I think we're OK as-is. The input is defined as a GitHub Actions choice, so the UI/API enforces it can only be one of the listed values. also we're only comparing the input value in expressions, not executing it in a shell, so there's no command injection path.

If we wanted to we could add an explicit validation step that fails unless the input is one of the expected values, but I don't think it's necessary given the current setup.

ChrisCookOC (approved):

DevarshBarotOC (approved):

PR #542, opened 2026-07-01
ci(LPF-1487): add workflow to scan for credentials

masuk-kazi98:
JIRA: [LPF-1487](https://dsdmoj.atlassian.net/browse/LPF-1487)

## Description of changes

Currently we have pre-commit hooks provided by hooks built by DevSecOps, that do credentials/secret scanning prior to commits. Only useful if developer has pre commit hooks enabled/installed. Even if enabled, commit hooks are easily bypassed using --no-verify as a commit argument. 

Took action by creating a Github workflow to enable secrets scanning on PRs but also full repo scanning on main + weekly scheduled scans, with the option of manual scans too.

.gitleaks.toml file added to specify allow-lists, false positives etc. Have currently added an ignore to UUIDs in text fixtures as it was picking these up from older commits

Githubs server side push protection should also block most secrets/credentials from getting through to remote

If a secret has been detected, dev must rebase to remove the affected commit

## Evidence
```
    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     add a test secret
password = REDACTED
Secret:      REDACTED
RuleID:      generic-api-key
Entropy:     5.000000
File:        README.md
Line:        481
Commit:      7ae170f95c826eee9044c4892854b70eb4314e97
Author:      Masuk Kazi
Email:       90158744+masuk-kazi98@users.noreply.github.com
Date:        2026-07-02T16:27:41Z
Fingerprint: 7ae170f95c826eee9044c4892854b70eb4314e97:README.md:generic-api-key:481

4:29PM INF 95 commits scanned.
4:29PM INF scan completed in 96.8ms
4:29PM WRN leaks found: 1
```

## Checklist
- [ ] Title follows the format `{type}({TICKET-NUMBER}): {brief description}`.   Example: `feat(ABC-123): add caching layer`
  - Type must be one of: 
    - feat
    - fix
    - docs
    - chore
    - refactor
    - test
    - ci
    - build
    - perf
- [ ] Clean commit history

[LPF-1487]: https://dsdmoj.atlassian.net/browse/LPF-1487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/secret-scanning.yml
```diff
@@ -0,0 +1,205 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ inputs.ref || github.event.pull_request.head.sha || github.sha }}
+          fetch-depth: 0
+
+      - name: Install gitleaks
+        run: |
+          curl -sSL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" -o gitleaks.tar.gz
```
DevarshBarotOC: Is there any reason we are installing gitleaks instead of using the official gitleaks action?

.github/workflows/secret-scanning.yml
```diff
@@ -0,0 +1,205 @@
...
+    steps:
+      - name: Checkout full history
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+          fetch-depth: 0
+
+      - name: Install gitleaks
+        run: |
+          curl -sSL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" -o gitleaks.tar.gz
```
DevarshBarotOC: Same with this, no real reason to download gitleaks unless we are specifically doing something the action doesn't already do for us.

.github/workflows/secret-scanning.yml
```diff
@@ -0,0 +1,205 @@
+name: Secret Scanning
```
DevarshBarotOC, line 1: As it currently stands and if my understanding is correct, if this runs on a weekly basis and a secret somehow makes/made it through, it will create the same issue every week as a duplicate until the issue is dealt with. I would suggest adding de-duplication logic in the script somewhere to ensure if a scan has picked up an issue previously, it does not create an issue for the same thing again.

.github/workflows/secret-scanning.yml
```diff
@@ -0,0 +1,205 @@
...
+          curl -sSL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" -o gitleaks.tar.gz
+          tar -xzf gitleaks.tar.gz gitleaks
+          sudo mv gitleaks /usr/local/bin/gitleaks
+          gitleaks version
+
+      - name: Determine commit range
+        id: range
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "log_opts=--all ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }}" >> "$GITHUB_OUTPUT"
```
DevarshBarotOC: For just a commit range, I would personally remove the `--all`  part. For PRs, you would only want the scan range to be `base..head` i.e. the commits from the end of the BASE (main) to the end of HEAD (your branch) because that is all that would've changed. 

Adding `--all` is asking git to start traversal from irrelevant places like 

```
refs/heads/*
refs/remotes/*
refs/tags/*
```

which will result in far more scans being done on irrelevant commits for just that PR. 

Tl;dr: We don't want to scan commits related to existing tags, existing active branches etc. for one PR every time someone raises one, so remove the `--all` flag.

.github/workflows/secret-scanning.yml
```diff
@@ -0,0 +1,205 @@
...
+              --log-opts="-1" \
+              --report-format sarif \
+              --report-path reports/gitleaks.sarif \
+              --redact \
+              --verbose \
+              --exit-code 1
+          fi
+
+      - name: Upload SARIF to Security tab
+        if: always()
```
DevarshBarotOC, line 132: We aren't checking the SARIF file exists here. Consider the following scenarios:
Happy path:
`We run a scan, gitleaks finds a problem, SARIF file created and uploaded, yay.`

Unhappy path (issue):
`We run a scan, something fails above (gitleaks download issue, corrupt checkout, someone introduces an invalid command line flag, syntax error, whatever). Now something fails, but we still run the upload with this line.`.

With the unhappy path, this upload will fail as the SARIF file will not exist as it never gets created, so now we have the following noisy log:
```
Run gitleaks -> nope
Upload SARIF -> woops, doesn't exist
Upload artifact -> nope because the last step failed
```

Basically it's worth checking the SARIF file exists before trying to upload it. If not, skip the upload.

DevarshBarotOC (changes requested):
Generally a pretty strong script, but there are a couple of small issues that are big enough to want to change it.

.github/workflows/secret-scanning.yml
masuk-kazi98: Fair question, i did initially trial out with the dedicated gitleaks action. One reason: gitleaks-action requires a GITLEAKS_LICENSE key for any repo owned by a GitHub organization. Since this is under ministryofjustice, we'd need a license to use the wrapper (easy to obtain). Not a blocker just a fyi. Secondary reason: using the CLI directly gives explicit control over diff-range vs full-history scanning rather than relying on the action's internal event-based heuristics

.github/workflows/secret-scanning.yml
masuk-kazi98: good point

.github/workflows/secret-scanning.yml
masuk-kazi98: yeah fairs, i agree

.github/workflows/secret-scanning.yml
DevarshBarotOC: Fair points, agreed licensing is overkill, and more importantly, we want that additional control which is great so nice one for catching that.

In this case, can you please amend this download to check the checksum against the archive? I suggest this because we are downloading and executing the binary, which we have no control over - i.e. it could be compromised - so a checksum verification would add that additional layer of verification that what we have downloaded is expected.

ChrisCookOC (approved):

DevarshBarotOC (approved):

H3ll3m4 (approved):
LGTM

PR #583, opened 2026-08-03
feat(LPF-1382): Initial integration with SDS (Secure Data Storage)

DevarshBarotOC:
JIRA: [LPF-1382](https://dsdmoj.atlassian.net/browse/LPF-1382)

## Description of changes
We wanted to spike out the code changes to integrate GPFD with SDS. This PR has the relevant changes that needed to be made to allow the Payforlegalaid application to communicate with SDS.

### Done:
- Added feature flag to control SDS feature availability in environments
    -  Controlled via env variables in Github, not config files, so we do not need to make config changes to enable the feature. 
- Added `/health`, `/get_file`, `/save_file` SDS endpoints via the SDS service
- Added a controller, token exchange service, SDS service and other wiring classes to ensure the communication with SDS occurs correctly.
- Completely separated current S3 Config from SDS config, so current flow remains unbroken and users can still authenticate with Entra and download files as expected.
- Added relevant exceptions and added them to global exception handler. 
- Updated various config (application-x.yml, deploy_api_to_env.yml, deployment.yml etc.) files to add new secrets/env variables which allow the application to communicate with the SDS client.

### To be completed:
- Clean up code and remove comments
- Static variables and comments where Sonar is is complaining.
- Add integration tests
- Add unit tests

## Evidence

| Endpoint | Evidence |
|----------|----------|
| `/get_file` endpoint working as expected in UI |  image   |
| `/health` endpoint working as expected in UI  | image |

## Checklist
- [X] Title follows the format `{type}({TICKET-NUMBER}): {brief description}`.   Example: `feat(ABC-123): add caching layer`
  - Type must be one of: 
    - feat
    - fix
    - docs
    - chore
    - refactor
    - test
    - ci
    - build
    - perf
- [ ] New tests are included if this a new feature
- [X] Tests and linter checks are passing locally
- [ ] Documentation README.md & Confluence have been updated
- [X] TODOs, commented code and print traces have been removed
- [X] Any dependant changes have been merged in downstream modules
- [X] Clean commit history

[LPF-1382]: https://dsdmoj.atlassian.net/browse/LPF-1382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #591, opened 2026-08-07
feat(LPF-1561)!: Migrate PFLA report retrieval to SDS

DevarshBarotOC:
JIRA: [LPF-1561](https://dsdmoj.atlassian.net/browse/LPF-1561)

## Description of changes
- Implementation to officially migrate PFLA to SDS.
    - Added openapi generator to use SDS' api docs to generate a contract.
    - Implemented service, controller and token related code to ensure SDS can communicate with the SDS client
    - Updated config where necessary. 

This PR was built on [PR 583(https://github.com/ministryofjustice/payforlegalaid/pull/583) that was created a part of the spike carried out for SDS ([LPF-1382](https://dsdmoj.atlassian.net/browse/LPF-1382)).

## Evidence

## Checklist
- [X] Title follows the format `{type}({TICKET-NUMBER}): {brief description}`.   Example: `feat(ABC-123): add caching layer`
  - Type must be one of: 
    - feat
    - fix
    - docs
    - chore
    - refactor
    - test
    - ci
    - build
    - perf
- [X] Tests and linter checks are passing locally
- [X] Documentation README.md & Confluence have been updated
- [X] TODOs, commented code and print traces have been removed
- [X] Any dependant changes have been merged in downstream modules
- [X] Clean commit history

[LPF-1561]: https://dsdmoj.atlassian.net/browse/LPF-1561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LPF-1382]: https://dsdmoj.atlassian.net/browse/LPF-1382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/uk/gov/laa/gpfd/config/builders/AuthorizeHttpRequestsBuilder.java
```diff
@@ -88,6 +98,8 @@ public void customize(AuthorizeHttpRequestsConfigurer<HttpSecurity>.Authorizatio
                 .requestMatchers(ACTUATOR_ENDPOINT, HEALTH_ENDPOINT).permitAll()         // Allow unrestricted access to actuator and health endpoints
                 .requestMatchers("/login").permitAll()
                 .requestMatchers(HttpMethod.POST, "/csp-report").permitAll()
+                .requestMatchers(SDS_HEALTH_ENDPOINT).permitAll()                        // Allow unrestricted access to SDS health endpoint
+                .requestMatchers(SDS_ENDPOINT).access(authManager)                       // SDS endpoints require authentication
```
ChrisCookOC, line 102: It doesn't harm but considering the next line is it necessary.

src/main/java/uk/gov/laa/gpfd/config/LocalAzureConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ * which is unavailable inside Docker, while still obtaining real tokens for calling
+ * the SDS test environment.
+ */
+@Slf4j
+@Configuration
+@Profile("local")
+@ConditionalOnProperty(name = "gpfd.sds-enabled.enabled", havingValue = "true")
+public class LocalAzureConfig {
+
+    @Value("${SDS_TENANT_ID:${AZURE_TENANT_ID:${AUTH_TENANT_ID:${TENANT_ID}}}}")
```
ChrisCookOC, line 26: So depeneding on the intention here depends on my suggestion. 

If we expect SDS and SILAS are sharing the same tenant id here, then lets just make this `$TENANT_ID` 

If we want to allow for them to be different, would suggest we just go for try `$SDS_TENANT_ID`, else `$TENANT_ID`

Note that running locally, using docker compose, the way things stand the first three of these will never be defined as they aren't passed into the container?

src/main/java/uk/gov/laa/gpfd/controller/SdsController.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * <p>Example usage:
+     * <pre>
+     *     GET /sds/files/CCMS%20AP%20Debtors.csv
+     * </pre>
+     *
+     * @param fileKey the name of the file to retrieve from SDS
+     * @return the download response containing the file URL
+     */
+    @GetMapping("/files/{fileKey}")
```
ChrisCookOC, line 42: Do these still need to be set up as controller accessible mappings when we just call `filesApi.getFile` directly etc.

src/main/resources/application-dev.yml
```diff
@@ -1,9 +1,15 @@
+logging:
+  level:
+    org.springframework.security: DEBUG
+    org.springframework.security.oauth2: TRACE
```
ChrisCookOC, line 4: are these needed in final version?

DevarshBarotOC: Not in prod, but seeing as this is the application-dev file, it doesn't hurt to have debug and trace turned on for OAuth2.0 - especially as token exchange and auth can be a bit of a pain to debug at times. Happy to remove if you feel strongly around not having debug logs in dev 😄

src/main/java/uk/gov/laa/gpfd/config/builders/AuthorizeHttpRequestsBuilder.java
DevarshBarotOC: It isn't necessary in terms of behaviour, but in my opinion it is in terms of intent. I guess it's a matter of "do we rely on a generic catch-all for endpoints that require auth, or do we explicitly state them?" - I know the current implementation uses the catch-all, but arguably it isn't "best practice". The way it is done here is easier to audit ("what auth rules apply to SDS?"), easier to modify later if needed, and reduces accidental config if things change.

src/main/java/uk/gov/laa/gpfd/controller/SdsController.java
DevarshBarotOC: Good point - I can get rid of the controller entirely, but do we wan to keep it for debug/development? I can mark it as such if so, or remove entirely - make the call, Cookie.

ChrisCookOC: If we keep it would like us to turn it off in prod. 

How much easier is it to debug via this than say the bruno collection you were playing with?

src/main/resources/application-dev.yml
ChrisCookOC: Debug fine, not sure about trace though :)

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -133,8 +143,21 @@ public ResponseEntity<StreamingResponseBody> getReportDownloadById(UUID id) {
...
+
+        var sdsService = sdsServiceProvider.getIfAvailable();
+        var sdsDownloadService = sdsDownloadServiceProvider.getIfAvailable();
+        if (sdsService != null && sdsDownloadService != null) {
+            try {
+                return streamFromSdsDownload(report, sdsService, sdsDownloadService);
+            } catch (SdsFileNotFoundException e) {
+                log.info("Report {} not found in SDS, falling back to current S3 download path", id);
+            }
+        }
```
masuk-kazi98, line 157: SdsService also throws ServiceUnavailableException on 500/502/503/504 from SDS, and SdsDownloadService throws it on download failure, neither is caught here. So if SDS is enabled but the service itself is degraded (not just missing the file), users get a hard error instead of falling back to S3, even though the file is still sitting in the S3 bucket.

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -2,30 +2,38 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Comparator;
+import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 
-import static uk.gov.laa.gpfd.model.FileExtension.CSV;
-import static uk.gov.laa.gpfd.model.FileExtension.S3STORAGE;
-import static uk.gov.laa.gpfd.model.FileExtension.XLSX;
+import static uk.gov.laa.gpfd.model.FileExtension.*;
```
masuk-kazi98, line 36: switching to wildcard imports now are we

src/main/java/uk/gov/laa/gpfd/services/sds/SdsTokenService.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.laa.gpfd.services.sds;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.stereotype.Service;
+import uk.gov.laa.gpfd.security.TokenProvider;
+
+import static java.time.Instant.*;
```
masuk-kazi98, line 9: switching to wildcard imports now are we

src/main/java/uk/gov/laa/gpfd/config/TokenCacheConfig.java
```diff
@@ -0,0 +1,12 @@
...
+import org.springframework.cache.CacheManager;
+import org.springframework.cache.concurrent.ConcurrentMapCacheManager;
+import org.springframework.context.annotation.Bean;
+
+public class TokenCacheConfig {
+    @Bean
+    CacheManager cacheManager() {
+        return new ConcurrentMapCacheManager("tokenCache");
+    }
+}
```
masuk-kazi98, line 12: TokenCacheConfig is not registered with Spring and the application has no @EnableCaching, so the cache annotations on TokenProvider are inert. the SDS-enabled IT mocks the services before token retrieval, so this is not covered. Either register/enable the cache and add a focused token-cache test, or remove this custom cache and rely on OAuth2AuthorizedClientManager's client lifecycle

src/main/java/uk/gov/laa/gpfd/controller/ReportsController.java
```diff
@@ -148,9 +171,61 @@ private ResponseEntity<StreamingResponseBody> fetchCsvExcelDownloadResponse(UUID
...
+                .stream()
+                .filter(detail -> detail.lastModified() != null)
+                .max(Comparator.comparing(SdsFileVersionDetail::lastModified));
+
+        return latestFileDetail.orElseThrow(() -> new SdsFileNotFoundException("File not found: " + fileKey));
+    }
+
+    private String buildSdsFilename(String reportName, SdsFileVersionDetail latestFileDetail) {
+        var uploadDate = latestFileDetail.lastModified().toLocalDate().format(DateTimeFormatter.ISO_LOCAL_DATE);
+        return "%s_%s.csv".formatted(reportName, uploadDate);
```
H3ll3m4, line 223: It is necessarily csv then

H3ll3m4:
This is great work but a big PR. We could have split in smaller PRs, like related to the tickets and the commit history could do with a bit of clean up. But it wasn't too bad because we reviewed the first ticket previously

